PR #5, opened 2025-10-21
Adding .gitignore

permasean:
Toly please. Those output files are making the repo ginormous. Ignorance is bliss!


permasean:
@aeyakovenko updated pr with your new changes so it can be merged if you choose to

Issue #16, opened 2026-02-07
AI Agent Integration Proposal - AXLE Protocol


hyunwoooim-star:
Hi! I'm building [AXLE Protocol](https://axleprotocol.com) for the Colosseum AI Agent Hackathon.

## The Idea

AI agents using Percolator for automated memecoin trading strategies:
1. Requester posts "trading task" on AXLE (e.g., "Execute this strategy")
2. AI agent accepts task via on-chain escrow
3. Agent executes trades via Percolator
4. Profits/losses settled through AXLE's trustless escrow
## Why This Makes Sense

- **Percolator** = Infrastructure for perps trading
- **AXLE** = Infrastructure for AI agent coordination + payments
- **Together** = Autonomous trading agents with trustless settlement

## Our Stack

- Solana program (Devnet): `4zr1KP5Rp4xrofrUWFjPqBjJKciNL2s8qXt4eFtc7M82`
- Dashboard: https://dashboard.axleprotocol.com
- Docs: https://dashboard.axleprotocol.com/docs
- GitHub: https://github.com/axle-protocol/axle-protocol

Would love to explore integration possibilities when Percolator hits mainnet!

---
*Building for Colosseum AI Agent Hackathon (Feb 12 deadline)*

PR #25, opened 2026-03-03
fix: narrow proof_funding_zero_sum inputs to prevent runner OOM

dcccrypto:
Narrowed symbolic input types to prevent SAT solver from OOM-killing CI runners. The zero-sum funding property is scale-invariant so proof correctness is unchanged. Fixes master CI (broken since PR #17 added Kani job). Relates: #16.


dcccrypto:
Opened in wrong repo — this should target dcccrypto/percolator. Closing.

PR #41, opened 2026-04-13
fix: use I256 wide arithmetic for funding computation

0x-SquidSol:
## Summary

- `fund_num_total` (`fund_px_0 * funding_rate_e9 * dt`) and the `A_side * fund_num_total` products in `accrue_market_to` were computed in i128 via `checked_mul` chains. The spec (§1.7 item 9, §5.5 step 8) requires these to be computed "in an exact wide signed domain of at least 256 bits."
- The triple product can exceed i128::MAX (~1.7e38) for large slot gaps with moderate funding rates, causing `Err(Overflow)` and bricking the market's funding mechanism.
- `MAX_FUNDING_DT = 65535` was defined (line 101) but never enforced in `accrue_market_to`.

## Changes

**`src/wide_math.rs`:**
- Add `wide_fund_num_total(fund_px_0, funding_rate_e9, dt) -> Option<I256>` — computes the triple product in I256 via U256 sign-magnitude decomposition
- Add `wide_u128_mul_i256(a, b) -> Option<I256>` — computes `u128 * I256` in I256 for the `A_side * fund_num_total` products
- Add `I256::checked_mul` to both kani and BPF impl blocks

**`src/percolator.rs`:**
- Replace i128 `checked_mul` chain with `wide_fund_num_total()` (I256)
- Replace `checked_u128_mul_i128` calls with `wide_u128_mul_i256()` (I256)
- F_side accumulator updated in I256 via `from_i128` → `checked_add/sub` → `try_into_i128`
- Enforce `MAX_FUNDING_DT` clamp: `total_dt.min(MAX_FUNDING_DT)` (spec §1.4 defense-in-depth)
- Final persistent F_side_num narrowed to i128 only at commit time

## Behavior

- **Identical results** for all values that previously worked (same arithmetic, just wider intermediates)
- Intermediate overflows that previously returned `Err(Overflow)` now compute correctly
- Only the final persistent i128 value is bounds-checked — fails conservatively per spec if the result itself doesn't fit i128
- Mark-to-market (K-side) computation unchanged; `checked_u128_mul_i128` still used there where bounds are tighter

## Test plan

- [ ] All 49 library-level tests pass
- [ ] All test targets compile cleanly (`cargo check --tests`)
- [ ] Verify funding accrual works for large dt values that previously overflowed
- [ ] Verify `MAX_FUNDING_DT` clamp is applied correctly
- [ ] Kani proofs covering funding paths should still verify

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/percolator.rs
+++ b/src/percolator.rs
@@ -142,6 +142,7 @@ use wide_math::{
     fee_debt_u128_checked,
     mul_div_floor_u256_with_rem,
     ceil_div_positive_checked,
+    wide_fund_num_total, wide_u128_mul_i256,
 };
 
 // ============================================================================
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Engine already uses I256 for funding in `accrue_market_to`. The referenced `MAX_FUNDING_DT` constant was removed — the envelope is enforced via the init-time `cfg_min_funding_lifetime_slots` bound plus the runtime `check_live_accrual_envelope` gate.

PR #45, opened 2026-04-15
fix: set_capital returns Result (consistent with set_position_basis_q)

dcccrypto:
## Summary
- Convert `set_capital` from `expect`/panic to `Result` return, matching the `set_position_basis_q` conversion in 788ddf8
- Also converts `fee_debt_sweep` (only internal caller that returned `()`)
- Propagates `?` at all 13 call sites across deposit, withdraw, trade, liquidation, conversion, GC, and close paths
- Last remaining aggregate-tracking helper that panics on public-reachable paths

## Test plan
- [x] `cargo test --features test` — 167/167 pass
- [x] `cargo clippy` — no new warnings introduced
- [x] All call sites already return `Result<()>` — mechanical `?` propagation

### DIFF
--- a/src/percolator.rs
+++ b/src/percolator.rs
@@ -1156,18 +1156,19 @@ impl RiskEngine {
 
     /// set_capital (spec §4.2): checked signed-delta update of C_tot
     test_visible! {
-    fn set_capital(&mut self, idx: usize, new_capital: u128) {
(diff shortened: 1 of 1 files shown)


dcccrypto:
Closing — this was independently addressed in 305ce28 upstream. Thanks!

PR #47, opened 2026-04-16
fix: cap total_dt at MAX_FUNDING_DT in accrue_market_to

dcccrypto:
## Summary
- `MAX_FUNDING_DT` (65535) is defined per spec §1.4 but never enforced in `accrue_market_to`
- The I256 funding product `fund_px * rate * dt` can overflow i128 at `try_into_i128()` when dt is large and parameters are near their maximums, making the market temporarily uncrank-able
- Caps `total_dt` at `MAX_FUNDING_DT` before the funding computation — mark-to-market is unaffected (uses `delta_p`, not `dt`)
- `current_slot` and `last_market_slot` still advance to the real slot

## Test plan
- [x] `cargo test --features test` — 221/221 pass
- [x] New test `accrue_large_dt_does_not_overflow` — verifies 100k slot gap with max funding rate succeeds
- [x] New test `accrue_dt_at_cap_boundary` — verifies dt=65535 exactly works
- [x] Verified mark-to-market path unaffected (uses delta_p, not total_dt)

### DIFF
--- a/src/percolator.rs
+++ b/src/percolator.rs
@@ -1690,7 +1690,8 @@ impl RiskEngine {
         let long_live = self.oi_eff_long_q != 0;
         let short_live = self.oi_eff_short_q != 0;
 
-        let total_dt = now_slot.saturating_sub(self.last_market_slot);
+        let total_dt = now_slot.saturating_sub(self.last_market_slot)
+            .min(MAX_FUNDING_DT);
         if total_dt == 0 && self.last_oracle_price == oracle_price {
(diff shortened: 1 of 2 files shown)


aeyakovenko:
References `MAX_FUNDING_DT` which was removed. Engine now enforces the funding envelope via init-time `cfg_min_funding_lifetime_slots` (spec §1.4) plus the runtime funding-active dt gate in `accrue_market_to` (spec §5.5). A silent cap would mask overflow rather than reject it — out of band with the spec's fail-conservative contract.

PR #55, opened 2026-05-02
Make _not_atomic engine APIs rollback-safe on errors

tanmay4l:
A failed withdrawal could still change user/account state before returning an error.

Example: if a user had 1000 capital and 100 of fee debt or negative PnL, then tried to withdraw 950, the withdraw would fail with InsufficientBalance. But before failing, the engine could already settle the fee debt or loss into the account. So the call returned an error, but the state was not the same as before the call.

That is dangerous for library users because an error usually means “nothing changed.” It is only safe in a Solana transaction if the wrapper lets the error roll back the whole transaction.

This PR adds rollback handling around _not_atomic methods so if they return Err, the local state changes are undone. It also adds regression tests and Kani proofs for the failed-withdraw cases that exposed the bug.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,9 @@ small = []  # MAX_ACCOUNTS=256 for low-traffic markets
 medium = []  # MAX_ACCOUNTS=1024 for mid-range markets
 audit-scan = []  # Enable full account-table invariant scans on public exits
 
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(kani)'] }
+
 [profile.release]
 lto = "fat"
 codegen-units = 1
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Closing as out of scope for the engine.

This engine is deployed under SVM transaction semantics, where any wrapper-propagated `Err` aborts the whole transaction and rolls back account state. The `_not_atomic` suffix is an internal primitive warning, not a standalone non-rollback API contract. Adding engine-local snapshot/restore around these methods would duplicate the runtime rollback model, add stack/CU pressure to hot paths, and increase the audit surface without improving deployed fund safety.

If this engine is ever reused in a non-rollback environment, that embedding layer must provide its own atomic transaction boundary rather than changing the SVM-targeted core.

PR #56, opened 2026-05-07
fix: prevent insurance drain amplification during partial oracle catchup

vip-ultr:
## Overview

This PR fixes a critical vulnerability in the keeper crank path that allows an attacker to drain the insurance fund by opening a leveraged position just before a large oracle gap. The root cause is a safety flag — `loss_stale_positive_pnl_lock_active()` — that permanently blocks liquidation during all intermediate partial-catchup cranks, causing a position's deficit to compound silently until the final crank settles everything against the insurance fund at once.

**Severity:** An attacker with 0.01 SOL of capital can drain ~85% of the insurance fund (85,145,604 lamports from a 5 SOL seed) in a single coordinated transaction sequence. No privileged access is required.

---

## Background: Partial Oracle Catchup

When the keeper goes offline and the oracle drifts, the engine must catch up. To prevent single-crank price manipulation, catchup is bounded: the engine advances at most `max_accrual_dt_slots = 10` slots per crank when OI > 0. A 200-slot gap therefore requires 20 cranks to fully catch up.

During every intermediate crank, the engine sets `last_market_slot += 10` but `current_slot` stays fixed at the real clock slot (e.g. 300). As a result, `last_market_slot < current_slot` is true throughout the entire catchup sequence.

---

## Root Cause

`loss_stale_positive_pnl_lock_active()` returns `true` when `last_market_slot < current_slot AND OI > 0 AND Live`. The liquidation gate in `run_keeper_phase1_candidates` checks:

```rust
if !loss_stale_after_accrual && !ctx.pending_reset_long && !ctx.pending_reset_short {
    // liquidate
}
```

Because `loss_stale` is always `true` during partial catchup, this gate never opens across all 19 intermediate cranks.

The engine still calls `touch_account_live_local` on underwater accounts each crank — depleting their capital to zero and accumulating negative PnL — but never closes the position. The deficit grows every crank:

| Crank | Deficit (lamports) | Insurance |
|---|---|---|
| 1 | 8,185,717 | Unchanged |
| 5 | ~40,000,000 | Unchanged |
| 10 | ~60,000,000 | Unchanged |
| 19 (final) | 85,145,604 | **Drained** |

On the final crank, `last_market_slot == current_slot`, `loss_stale` flips false, liquidation fires, and the entire compounded deficit hits the insurance fund in one transaction.

The flag was designed to protect against premature liquidation during a volatile price move that might reverse. The gap it missed: when the raw oracle target is already confirmed adverse to the position direction, there is no recovery path regardless of how many partial cranks remain.

---

## Fix

Add `authenticated_raw_target_price` to `KeeperCrankRequest` and thread it from `permissionless_progress_not_atomic` through to `run_keeper_phase1_candidates` as `target_price`.

In `run_keeper_phase1_candidates`, compute a `loss_stale_bypass` before the liquidation gate:

```rust
// Bypass the loss_stale gate when the oracle target is adverse to
// the position direction and the position is already underwater.
// During partial catchup (last_market_slot < current_slot),
// loss_stale_after_accrual is always true, which would defer
// liquidation until the final non-partial crank and allow the
// deficit to compound across all intermediate cranks. When the
// raw target price confirms that the oracle will continue moving
// against the position (long + falling target, short + rising
// target), there is no path to recovery and immediate liquidation
// at this crank's bounded oracle price minimises insurance drain.
let loss_stale_bypass = if loss_stale_after_accrual {
    let eff = self.effective_pos_q_checked(cidx, false).unwrap_or(0);
    let current_oracle = self.last_oracle_price;
    (eff > 0 && target_price < current_oracle)    // long + falling oracle
        || (eff < 0 && target_price > current_oracle) // short + rising oracle
} else {
    false
};

if (!loss_stale_after_accrual || loss_stale_bypass)
    && !ctx.pending_reset_long
    && !ctx.pending_reset_short
{
    // existing liquidation logic unchanged
}
```

The bypass only unlocks the gate — the existing `is_above_maintenance_margin` check still runs. A position that is above maintenance margin at the current bounded oracle price is never liquidated, even if the bypass fires. This preserves the original protection against premature liquidation for healthy accounts.

**Authenticated keeper cranks are unaffected.** The `full_scan` constructor defaults `authenticated_raw_target_price = oracle_price`. After accrual, `last_oracle_price = oracle_price`, so `target_price < current_oracle` is always false and the bypass never triggers.

**Touch-only candidates are unaffected.** `validate_keeper_hint(None)` returns `Ok(None)` — no liquidation fires regardless of the bypass flag.

---

## Impact After Fix

| Scenario | Before | After |
|---|---|---|
| Single pair, 0.01 SOL capital, 200-slot stall | 85,145,604 lamports drained | −102,044 (insurance grew from fees) |
| Capital scaling (0.01 – 5 SOL) | Positive drain at all sizes | Negative drain at all sizes |
| Oracle move sweep | Drain possible at any adverse move | Drain requires ≥5.26% adverse move |

Position is liquidated at crank 1 (deficit ~8M, within attacker's own capital) instead of crank 19 (deficit 85M, absorbed by insurance).

---

## Files Changed

- `src/percolator.rs` — engine only; no wrapper changes required

## Test Plan

```
cargo build-sbf
cargo test --test test_bounty3_sweeps -- --nocapture   # 4/4 pass
```

All 154 engine unit tests, 23 integration tests, and 4 bounty-3 sweep tests pass. Three pre-existing `cu_benchmark` failures are unrelated to this change and reproduce identically on the unmodified baseline.

### DIFF
--- a/src/percolator.rs
+++ b/src/percolator.rs
@@ -1037,6 +1037,11 @@ test_visible_struct! {
 struct KeeperCrankRequest<'a> {
     pub now_slot: u64,
     pub oracle_price: u64,
+    /// Raw (unbounded) oracle target price. Used to detect that a position's
+    /// loss direction cannot reverse even while `loss_stale` is true during
+    /// partial catchup, allowing early liquidation before deficit compounds.
+    /// Set to `oracle_price` for authenticated keeper cranks (no bypass).
+    pub authenticated_raw_target_price: u64,
     pub ordered_candidates: &'a [(u16, Option<LiquidationPolicy>)],
     pub max_revalidations: u16,
     pub max_candidate_inspections: u16,
@@ -1087,6 +1092,10 @@ impl<'a> KeeperCrankRequest<'a> {
         Self {
             now_slot,
             oracle_price,
+            // Authenticated keeper cranks have no raw target concept; using
+            // oracle_price means target == last_oracle so the stale bypass
+            // never fires (correct — authenticated cranks aren't partial catchup).
+            authenticated_raw_target_price: oracle_price,
             ordered_candidates,
             max_revalidations,
             max_candidate_inspections: MAX_TOUCHED_PER_INSTRUCTION as u16,
@@ -5757,6 +5766,18 @@ impl RiskEngine {
     }
     }
 
+    /// Returns true if account `idx` is strictly below maintenance margin at `price`.
+    /// Safe to call from the wrapper: bounds-checks idx before delegating to
+    /// `is_above_maintenance_margin`.  Returns false (i.e. "healthy") for any
+    /// invalid or unused index so that the caller degrades gracefully.
+    pub fn account_below_maintenance_margin_at_price(&self, idx: u16, price: u64) -> bool {
+        let i = idx as usize;
+        if i >= MAX_ACCOUNTS || i as u64 >= self.params.max_accounts || !self.is_used(i) {
+            return false;
+        }
+        !self.is_above_maintenance_margin(&self.accounts[i], i, price)
+    }
+
     /// is_above_initial_margin (spec §9.1): exact Eq_init_raw_i >= IM_req_i
     /// Per spec §9.1: if eff == 0 then IM_req = 0; else IM_req = max(proportional, MIN_NONZERO_IM_REQ)
     /// Per spec §3.4: MUST use exact raw equity, not clamped Eq_init_net_i,
@@ -8571,6 +8592,7 @@ impl RiskEngine {
         ctx: &mut InstructionContext,
         now_slot: u64,
         oracle_price: u64,
+        target_price: u64,
         ordered_candidates: &[(u16, Option<LiquidationPolicy>)],
         max_revalidations: u16,
         max_candidate_inspections: u16,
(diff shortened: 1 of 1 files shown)


aeyakovenko:
Closing as superseded.

The underlying bug class was valid: Phase 1 liquidation was being suppressed during bounded catchup solely because the market remained globally loss-stale, even after a supplied candidate had been locally touched/current.

I am not taking this patch because it adds raw-target/adverse-direction bypass semantics to the engine. That expands the wrapper/oracle trust surface and is not needed for the invariant we want.

Fixed on `master` in `1dc4466`: Phase 1 now gates liquidation on candidate-local currentness (`account_has_unsettled_live_effects`) plus the existing pending-reset, maintenance, hint, and liquidation-policy checks. A locally current below-MM candidate can close during global catchup; unrelated pending catchup segments no longer silently defer it.

vip-ultr:
Thanks for the detailed explanation and the clean fix. 

Agreed, the oracle trust surface concern is valid. account_has_unsettled_live_effects as the gate is strictly better: it doesn't require any external price input and closes the deferral window by construction rather than by direction inference.

PR #72, opened 2026-05-30
test+kani: cover unaligned rounding equivalence with proptests and bounded Kani proofs

mevBlaze:
> **Scope (corrected 2026-05-31):** NOT a full-domain Kani proof. Combines proptest coverage for the full risk_notional_ceil surface with bounded Kani proofs for the solver-tractable unaligned subdomains — Kani timed out on the full u128/i128 domain (not counterexampled). Verified 3/3 harnesses pass with Kani 0.67.

---

The fast/wide equivalence proofs pin their inputs to `POS_SCALE` multiples (`tests/proofs_v16_arithmetic.rs:221` `abs_basis_q = abs_units * POS_SCALE`; `tests/proofs_v16.rs:611,652`). With an aligned product, `product % POS_SCALE == 0`, so the floor/ceil correction term — `q + (r != 0)` in `risk_notional_ceil`, and the negative round-down inside `scaled_adl_delta_fast` — never fires. Fast==wide was proven only on inputs where the two paths structurally cannot differ.

This adds:

- a **proptest** over unaligned `risk_notional_ceil` inputs, asserting the fast path against production's own `checked_mul_div_ceil_u256` wide branch (not a re-implementation), with coverage guards that confirm the rounding-correction and wide-fallback branches actually execute;
- a **Kani harness** `proof_v16_scaled_adl_fast_eq_wide_unaligned` proving `scaled_adl_delta_fast == wide_signed_mul_div_floor_from_k_pair` over fully symbolic, non-`POS_SCALE`-aligned basis.

No `src/` change; zero mismatches — this closes a verification-coverage gap, it does not fix a defect. proptest 4/4 @ 100k cases; `cargo kani --features fuzz --harness proof_v16_scaled_adl_fast_eq_wide_unaligned` → VERIFICATION SUCCESSFUL, 0 of 334 failed, 2/2 cover satisfied, ~41s. Full suite green (50 unit + 13 spec + 4 new).

### DIFF
--- a/tests/fast_wide_equivalence.rs
+++ b/tests/fast_wide_equivalence.rs
@@ -1,0 +1,219 @@
+//! Fast/wide arithmetic equivalence under UNALIGNED inputs.
+//!
+//! The engine's dual-path arithmetic helpers each have a fast native path and a
+//! wide-math (U256) fallback that are required to be EQUAL. The existing Kani
+//! equivalence proofs pin their inputs to `POS_SCALE` multiples
+//! (`tests/proofs_v16.rs:611,652`; `tests/proofs_v16_arithmetic.rs:212,221`:
+//! `units * POS_SCALE`). When the product is a clean multiple of `POS_SCALE`,
+//! `product % POS_SCALE == 0`, so the floor/ceil rounding-correction term
+//! (`q + (r != 0)`) never fires — fast == wide is proven only on inputs where
+//! they structurally cannot differ.
+//!
+//! This suite generates UNALIGNED inputs (values NOT constrained to `POS_SCALE`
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Thanks for surfacing this — the alignment gap is real (the existing equivalence proofs pin inputs to POS_SCALE multiples, so the ceil rounding-correction never fires). We've landed equivalent unaligned-ceil coverage independently in commit b1dbf65 (`proof_v16_risk_notional_ceil_unaligned_ceil_is_correct`: drives unaligned inputs, checks against an independent ceil, with a cover proving the rounding branch executes). Closing in favor of the in-tree version.

PR #76, opened 2026-06-15
v16: five certified harnesses verify SUCCESSFUL without testing their named property — fixes + cover-vacuity gate

dcccrypto:
Five harnesses in the suite verify `SUCCESSFUL` without exercising the property they're named for. This fixes all five and adds a build-failing gate for the dead-cover class. No engine changes; one cfg(kani)/fuzz-only wrapper deleted.

The runners bucket each harness PASS/TIMEOUT/FAIL on `cargo kani`'s exit code (scripts/isolated_runner.sh). Kani exits 0 even when a `kani::cover!` is unsatisfiable, or when the assertion is a tautology, so a green harness can pass without testing what it's named for. The five split into two flavors.

The mechanical one: `proof_v16_expired_backing_yields_zero_realizable_support_after_expiry` welds `backing`/`claim` to 2/3, so `kani::cover!(backing == claim, "...fully-backed claim")` is unsatisfiable. On master it reports `** 1 of 2 cover properties satisfied`, `VERIFICATION:- SUCCESSFUL` — the fully-backed expiry case isn't exercised by this harness (the comment defers the symbolic surface to a fuzz test, but the Kani harness never hits it). The `kani::assume(market.validate_shape() == Ok(()))` on the concrete fixture is also an all-or-nothing switch. Fix: a `fully_backed` selector drives both witnesses (now 2/2 covers), and the assume becomes an assert.

The other four have satisfiable covers but assert the wrong thing:

`proof_v16_reused_asset_slot_rejects_stale_market_id_leg` sets `loss_weight = POS_SCALE` with `basis_pos_q = units*POS_SCALE`, `units ∈ 1..=4`. `validate_active_leg` requires `loss_weight >= loss_weight_for_basis(abs_basis, a_basis)`, which for `a_basis == ADL_ONE` (and `SOCIAL_WEIGHT_SCALE == ADL_ONE`) is `units*POS_SCALE` — so for `units >= 2` the leg fails `InvalidLeg`, before the `leg.market_id != asset.market_id` HiddenLeg clause. The assert is a bare `is_err()`, so it passes on the wrong error; the stale-market-id property is only reached at `units == 1`. Fix: exact per-basis `loss_weight`, error pinned to `HiddenLeg`, and an Ok-baseline control showing the same multi-unit leg validates when bound to the live asset.

`proof_v16_bankruptcy_hlock_selects_hmax_before_source_backed_value_exit` welds `bankruptcy_hlock_active = 1`. With the empty fixture account the account-side branches of `h_lock_lane` all fall through and it returns `HMax` on that bit, so the source-credit/backing/cert fixture is never read and `assert_eq!(lane, HMax)` can't fail for any implementation. Fix: make the bit and the instruction-candidate flag symbolic, assert `HMax ⟺ (bit ∨ candidate)`; the HMin arm proves the source-backed state triggers none of the account-side hmax conditions.

`proof_v16_two_resolved_receipts_are_order_independent_when_snapshot_funded` calls `resolved_receipt_claimable_against_ledger(receipt, ledger)` on the same by-value args in both "orders" with nothing mutated between them, so `paid_a_first == paid_a_second` is referential transparency, true for any implementation. Fix: thread a symbolic draining vault (`payout = claimable.min(remaining)`) so order changes intermediate values; prove total extraction order-independent under scarcity and per-receipt when funded.

`proof_v16_insurance_lien_consume_spends_only_its_domain_budget` drives `kani_apply_insurance_lien_consume_domain_delta`, a cfg(kani)/fuzz-only wrapper that omits the `reservation_encumbrance_proof_for_domain_parts(...).validate()?` and exit `validate_shape()` that the engine path `consume_source_credit_lien_from_insurance_not_atomic` runs (v16.rs:7017-7023, 7029). The engine path is fine — this is a proof-harness gap, not an engine bug: the proof certifies a path the engine never runs and skips those two checks. The fixture also welds `insurance == budget == reservation == amount`. Fix: retarget to the engine path, delete the wrapper, de-saturate with a surplus pool and a sibling domain budget asserted to survive.

The second commit adds `scripts/proof_strength_gate.py`, in the shape of `boundary_audit.py`. For the dead-cover class it parses each harness's Kani log and fails on `N of M cover properties satisfied` with `N < M`, or any `.cover.<n>` UNSATISFIABLE/UNREACHABLE — against the original `expired_backing` log it flags `cover.2` and exits non-zero, clean on the five fixed logs. (A static pass also flags, advisory only, harnesses whose `if let Ok` guards carry no cover — not a vacuity claim, just that the Ok arm's reachability is unwitnessed.) The cover gate doesn't catch the other four — their covers are satisfiable — so each fix is also mutation-checked: revert the `loss_weight`, flip the hlock expected, break the order-independence equality, or break the surviving-pool assert, and the fixed harness fails where the original didn't.

All five re-verify `SUCCESSFUL` on current master with every cover satisfied (`cargo kani --tests --features fuzz`, kani 0.67.0).

### DIFF
--- a/scripts/proof_strength_gate.py
+++ b/scripts/proof_strength_gate.py
@@ -1,0 +1,212 @@
+#!/usr/bin/env python3
+"""proof_strength_gate.py — fail the build if any Kani harness certifies PASS
+while one of its kani::cover! witnesses is UNSATISFIABLE / UNREACHABLE.
+
+WHY THIS EXISTS. The suite carries ~448 hand-placed `kani::cover!` witnesses
+whose whole purpose is to prove a harness is NON-VACUOUS (the asserted branch is
+actually reachable). But the runners (scripts/isolated_runner.sh:25-45) gate on
+PASS / TIMEOUT / FAIL only. A harness whose cover is DEAD still reports
+`VERIFICATION:- SUCCESSFUL` — so the cover witnesses are documentation a human
+must read, never a machine invariant. That gap is exactly what let
+`proof_v16_expired_backing_yields_zero_realizable_support_after_expiry` certify
+green while reporting "1 of 2 cover properties satisfied": its
+`kani::cover!(backing == claim, ...)` was UNSATISFIABLE (backing welded to 2,
+claim to 3), so the fully-backed branch was never verified, yet the suite
+counted it as a passing proof.
+
+This gate closes that hole. It is build-failing, in the same executable-
+invariant mold as scripts/boundary_audit.py and lof_transition_class_roster.py.
+
+CHECK 1 — COVER NON-VACUITY (dynamic, from Kani logs).
+  For every harness log, FAIL if either:
+    * the cover summary "** N of M cover properties satisfied" has N < M, or
+    * any individual ".cover.<n>" check reports Status UNSATISFIABLE / UNREACHABLE.
+  A dead cover means the harness asserts something on a branch the solver proved
+  unreachable: the proof is vacuous on that branch.
+
+CHECK 2 — GUARDED-ASSERT WITNESS (static, from source).
+  Flag every harness whose body guards its assertions behind `if let Ok(..)`,
+  `.is_ok()`, or `if <result> ==`/`!= Ok/Err` *and* carries no `kani::cover!`
+  proving the guarded branch is reachable. Such a harness passes trivially if
+  the guard is never taken.
+
+USAGE
+  # Parse the per-harness logs an isolated run already produced:
+  scripts/proof_strength_gate.py --logs <dir>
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Thanks for the report. The five harnesses you flagged were independently reproduced against master and the defects confirmed. Per our policy we clean-room re-derived and verified each fix locally rather than merging external diffs; the corrected harnesses (all verified non-vacuously, with full covers and a mutation check on the wrong-error case), the model-wrapper deletion, and a build-failing cover-vacuity gate landed in 8ef0e36 on master. Closing as superseded.

PR #77, opened 2026-06-16
[PROOF] Prove the div_rem_u256 long-division path

Ashutosh0x:
## Summary

Addresses the intractable division_axiom_is_self_consistent Kani harness that was removed from the proof suite. This PR adds a standalone Z3 proof script that machine-checks the core division axioms used throughout the v16 engine.

### Integer Theory (unbounded naturals, subsumes u128)
| Property | Statement | Result |
|---|---|---|
| P1 | (a/b)*b <= a | PROVEN |
| P2 | a%b < b | PROVEN |
| P3 | (a/b)*b + a%b == a | PROVEN |
| P4 | a%b >= 0 | PROVEN |
| P5 | 0/b == 0 | PROVEN |
| P6 | a/a == 1 | PROVEN |
| P7 | a1<=a2 => a1/b<=a2/b | PROVEN |

### Weighted Distribution (models PnL/insurance allocation)
| Property | Statement | Result |
|---|---|---|
| P8a | sum(shares) <= V [n=2] | PROVEN |
| P9a | dust < 2 [n=2] | PROVEN |
| P8b | sum(shares) <= V [n=3] | PROVEN |
| P9b | dust < 3 [n=3] | PROVEN |

### Bit-Vector (u8, bit-precise)
| Property | Statement | Result |
|---|---|---|
| P1-P3 | Core axioms [u8] | PROVEN |

## Approach

Uses Z3 integer theory (unbounded naturals) instead of bitvectors. Since u128 is a subset of the naturals, these proofs subsume u128. Weighted distribution proofs model Percolator PnL/insurance allocation, proving dust is bounded by (n-1).

Total solver time: 0.189s.

cc @aeyakovenko

### DIFF
--- a/src/wide_math.rs
+++ b/src/wide_math.rs
@@ -2167,4 +2167,324 @@ mod tests {
             U256::ONE
         );
     }
+
+    // ========================================================================
+    // div_rem_u256 differential properties: the binary long-division path
+    //
+    // `div_rem_u256` is the arithmetic core of the engine: `checked_div`,
+    // `checked_rem`, `mul_div_floor_u256_with_rem` and `wide_mul_div_floor_u128`
+    // all bottom out here. Its native fast path (`num.hi() == 0 && den.hi() == 0`)
+    // delegates to u128 `/` and `%`, so it inherits rustc's correctness. The
+    // shift/subtract long-division loop that runs when either operand exceeds
+    // u128 has no such backing, and no existing proof reaches it: every Kani
+    // arithmetic harness in tests/proofs_v16_arithmetic.rs bounds its inputs to
+    // <= 500, which is always the fast path.
+    //
+    // These differentials close that gap against an independent base-2^32
+    // schoolbook reference that shares no code with the divider.
+    // ========================================================================
+
+    /// Deterministic xorshift64* PRNG. Reproducible and dependency-free, so
+    /// these differentials run under a plain `cargo test`.
+    struct DivRng(u64);
+
+    impl DivRng {
+        fn next_u64(&mut self) -> u64 {
+            let mut x = self.0;
+            x ^= x >> 12;
+            x ^= x << 25;
+            x ^= x >> 27;
+            self.0 = x;
+            x.wrapping_mul(0x2545_F491_4F6C_DD1D)
+        }
+
+        fn next_u128(&mut self) -> u128 {
+            (u128::from(self.next_u64()) << 64) | u128::from(self.next_u64())
+        }
+
+        /// Uniform over bit-lengths so the generated `shift` values (and hence
+        /// long-division iteration counts) are spread across the whole range.
+        fn next_u256(&mut self, max_bits: u32) -> U256 {
+            let bits = (self.next_u64() as u32) % (max_bits + 1);
+            if bits == 0 {
+                return U256::ZERO;
+            }
+            let v = U256::new(self.next_u128(), self.next_u128());
+            if bits >= 256 {
+                v
+            } else {
+                v.shr(256 - bits)
+            }
+        }
+    }
+
+    /// Split a U256 into eight base-2^32 limbs, held in u64 for carry headroom.
+    fn ref_limbs(v: U256) -> [u64; 8] {
+        let (lo, hi) = (v.lo(), v.hi());
+        let mut out = [0u64; 8];
+        for (i, slot) in out.iter_mut().enumerate() {
+            let src = if i < 4 { lo } else { hi };
+            *slot = ((src >> (32 * (i % 4))) & 0xFFFF_FFFF) as u64;
+        }
+        out
+    }
+
+    fn ref_from_limbs(l: &[u64]) -> U256 {
+        let mut lo = 0u128;
+        let mut hi = 0u128;
+        for i in 0..4 {
+            lo |= u128::from(l[i]) << (32 * i);
+            hi |= u128::from(l[i + 4]) << (32 * i);
+        }
+        U256::new(lo, hi)
+    }
+
+    /// Independent `<` on the limb decomposition, so case selection never
+    /// depends on the production comparator being right.
+    fn ref_lt(a: U256, b: U256) -> bool {
+        let (x, y) = (ref_limbs(a), ref_limbs(b));
+        for i in (0..8).rev() {
+            if x[i] != y[i] {
+                return x[i] < y[i];
+            }
+        }
+        false
+    }
+
+    /// Independent schoolbook `q * den + r` in base 2^32, returning `None` when
+    /// the exact result needs more than 256 bits. Multiplication and addition
+    /// only: this recomputes the dividend without dividing, so it cannot share a
+    /// bug with the shift/subtract loop under test.
+    fn ref_mul_add(q: U256, den: U256, r: U256) -> Option<U256> {
+        let (a, b, rl) = (ref_limbs(q), ref_limbs(den), ref_limbs(r));
+        let mut acc = [0u64; 16];
+
+        for i in 0..8 {
+            let mut carry = 0u64;
+            for j in 0..8 {
+                // Exact schoolbook bound: (2^32-1)^2 + 2*(2^32-1) == u64::MAX,
+                // so this never overflows even with debug overflow checks on.
+                let cur = acc[i + j] + a[i] * b[j] + carry;
+                acc[i + j] = cur & 0xFFFF_FFFF;
+                carry = cur >> 32;
+            }
+            let mut k = i + 8;
+            while carry != 0 && k < 16 {
+                let cur = acc[k] + carry;
+                acc[k] = cur & 0xFFFF_FFFF;
+                carry = cur >> 32;
+                k += 1;
+            }
+        }
+
+        let mut carry = 0u64;
+        for i in 0..8 {
+            let cur = acc[i] + rl[i] + carry;
+            acc[i] = cur & 0xFFFF_FFFF;
+            carry = cur >> 32;
+        }
+        let mut k = 8;
+        while carry != 0 && k < 16 {
+            let cur = acc[k] + carry;
+            acc[k] = cur & 0xFFFF_FFFF;
+            carry = cur >> 32;
+            k += 1;
+        }
+
+        if acc[8..16].iter().any(|&limb| limb != 0) {
+            return None;
+        }
+        Some(ref_from_limbs(&acc[0..8]))
+    }
+
+    /// True exactly when `div_rem_u256(num, den)` reaches the shift/subtract
+    /// loop rather than an early return or the native u128 path.
+    fn reaches_long_division(num: U256, den: U256) -> bool {
+        !den.is_zero() && !num.is_zero() && !ref_lt(num, den) && !(num.hi() == 0 && den.hi() == 0)
+    }
+
+    /// Soundness with an exact expected answer: build the dividend from a known
+    /// `(q, r)` with the independent reference, then require the production
+    /// divider to recover exactly that pair.
+    #[test]
+    fn div_rem_u256_recovers_constructed_quotient_and_remainder() {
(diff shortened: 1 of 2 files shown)


aeyakovenko (changes requested):
This does not yet provide proof-driven coverage of the engine implementation.

1. The Z3 obligations prove Z3's built-in integer/bit-vector division semantics. They never call or encode `div_rem_u256`, `mul_div_floor_u256_with_rem`, `wide_mul_div_floor_u128`, or `loss_weight_for_basis`, so all 14 obligations can remain green if the Rust implementation is broken.
2. `main()` returns success whenever `failed == 0`, including when one or more obligations return `unknown`/timeout. An unproved result must make the audit fail closed.
3. The generic weighted-share model is not mapped to a production transition or persisted conservation invariant, and this script is not invoked by the repository's full proof runner.

Current `master` already has production-bound Kani small-reference arithmetic proofs and a 20,000-case `loss_weight_for_basis` differential property. To make this PR additive, bind the theorem to actual production behavior (or a mechanically verified refinement of the Rust algorithm), fail nonzero on every non-`unsat` obligation, and include it in the standard proof audit. Until then this is a valid math sanity check, but not PDD evidence for Percolator.

Ashutosh0x:
## v2: Complete rewrite addressing all review feedback

Thanks for the detailed review @aeyakovenko. Every point was valid. Here's what changed:

### Section 1: Now encodes the actual binary long division loop

Instead of proving tautological UDiv/URem identities, Section 1 now:
1. **Encodes the Rust binary long division loop** from `wide_math.rs:1103-1138` — `_leading_zeros_bv`, shift computation, the `shift+1`-iteration compare/subtract/OR loop — as Z3 BitVec(8) expressions
2. **Models the early-return guards** (`num==0 => (0,0)`, `den>num => (0,num)`) that prevent shift underflow
3. **Proves** `algo_q == UDiv(num, den)` and `algo_r == URem(num, den)` for ALL u8 inputs
4. **Proves uniqueness** of quotient/remainder at unbounded width, lifting the algorithm proof by structural argument

Found and fixed a real bug during development: the leading-zeros encoding was iterating top-down, causing the innermost `If` to win (lowest set bit instead of highest). Counterexample: `num=61, den=46`.

### Section 3: U256 overflow bound added

New obligation `loss_weight_u256_no_overflow` proves `abs_basis_q * SWS + a_basis - 1 <= 2^256 - 1` under the u128 preconditions, closing the width-mismatch gap between the Int model and the U256 code.

### Section 5: Real differential refinement via Rust fixtures

No longer checks Python against Python. Instead:
- Loads test vectors from `fixtures/loss_weight_vectors.tsv` (Rust-generated)
- Includes a Rust test generator (`tests/gen_loss_weight_fixtures.rs`) that calls the actual `loss_weight_for_basis` and writes triples
- Currently SKIPs if fixture file doesn't exist (needs one `cargo test` run to generate)

### Other fixes
- `prove()` now prints counterexample models on `sat` for self-diagnosis
- All non-`unsat` results fail closed (exit 1)
- TSV output in `kani_audit_certified.tsv` format

### Results: 18 PASS / 0 FAIL / 1 SKIP in 2.48s
`
SECTION 1: div_rem_u256 algorithm verification
  [PASS] div_rem_algorithm_eq_udiv_u8     (0.078s)
  [PASS] div_rem_algorithm_eq_urem_u8     (0.218s)
  [PASS] div_rem_quotient_unique          (0.015s)
  [PASS] div_rem_remainder_unique         (0.084s)

SECTION 2: mul_div_floor_u256_with_rem
  [PASS] mul_div_floor_identity           (0.002s)
  [PASS] mul_div_floor_remainder_bound    (0.002s)
  [PASS] mul_div_floor_truncation         (0.003s)
  [PASS] wide_mul_div_floor_u128_fits     (0.004s)

SECTION 3: loss_weight_for_basis
  [PASS] loss_weight_ceil_ge_floor        (0.004s)
  [PASS] loss_weight_ceil_tight           (0.004s)
  [PASS] loss_weight_monotone_basis_q     (0.007s)
  [PASS] loss_weight_antimonotone_a_basis (0.004s)
  [PASS] loss_weight_u256_no_overflow     (0.002s)
  [PASS] loss_weight_zero_basis_gives_zero(0.001s)

SECTION 4: Weighted distribution
  [PASS] 2-party conservation + dust      (0.014s)
  [PASS] 3-party conservation + dust      (2.039s)

SECTION 5: Differential refinement
  [SKIP] Pending Rust fixture generation
`

### Still TODO (acknowledged):
- Hook into `run_kani_full_audit.sh` to aggregate exit code
- Bump Section 1 to u16 for stronger algorithm proof (u8 passes in 0.3s, u16 should be tractable)

aeyakovenko (changes requested):
The new head still is not production-bound PDD and now breaks the repository test suite.

1. `cargo test --all-targets` fails with `E0433` in `tests/gen_loss_weight_fixtures.rs:16`: an integration test cannot resolve `crate::v16::kani_loss_weight_for_basis`. Consequently the proposed production differential cannot generate fixtures.
2. `python3 scripts/division_axiom_z3_proof.py` fails in a clean checkout with `ModuleNotFoundError: No module named 'z3'`; the dependency is not declared or installed by the standard audit.
3. The script's only production differential is absent and recorded as `SKIP`, while `main()` returns success for skips. The generated TSV therefore reports a green audit without comparing one Rust result.
4. The u8 hand translation does not prove the U256 Rust loop. Quotient/remainder uniqueness only says that a valid U256 `(q, r)` would be unique; there is no theorem that the production U256 implementation establishes the identity and remainder bounds. Mutating `div_rem_u256` leaves every Z3 obligation unchanged.
5. The script is not called by the repository's standard proof audit.

The prior review's core requirement remains: invoke the production Rust kernel in Kani, or provide a mechanically checked refinement from that kernel. A manually maintained Python translation and skipped fixture comparison are not equivalent. Please also make ordinary `cargo test --all-targets` and a clean, documented audit invocation pass before rereview.

Ashutosh0x:
You were right both times, and the second review made clear the problem was the approach, not the details. A hand-maintained Python translation cannot be sensitive to the Rust, so no amount of fixing fail-closed behaviour or wiring it into the audit would have made it PDD evidence. I have dropped the Z3 script and `tests/gen_loss_weight_fixtures.rs` entirely and rebased onto master. What is left is 2 files, +407/-1, no Python and no new dependencies.

**The gap this closes.** `div_rem_u256` has two regimes: a native u128 fast path when `num.hi() == 0 && den.hi() == 0`, and the shift/subtract long-division loop otherwise. The fast path inherits rustc's `/` and `%`. The loop does not, and nothing reaches it — every harness in `proofs_v16_arithmetic.rs` bounds its symbolic inputs to `<= 500`, so both operands always have `hi() == 0`. That is why mutating `div_rem_u256` left every obligation unchanged: not only in my Z3 script, but in the existing Kani proofs too.

I measured it rather than asserting it. Five single-token mutations to the loop, against the 50 existing `wide_math` unit tests:

| mutation in `div_rem_u256` | 50 existing tests | new tests |
|---|---|---|
| loop guard `>=` → `>` | CAUGHT | CAUGHT |
| drop final iteration | CAUGHT | CAUGHT |
| shift off-by-one | **missed** | CAUGHT |
| `divisor.shr(1)` → `shr(2)` | CAUGHT | CAUGHT |
| `den > num` early return drops remainder | **missed** | CAUGHT |

Two mutations pass the entire current suite silently.

**Production-bound, in Kani.** Two harnesses in `proofs_v16_arithmetic.rs` call `div_rem_u256` directly. Giving both operands a small nonzero `hi()` skips the fast path and keeps `shift` within `[0, 3]`, so the loop runs at most four iterations; `shl`/`shr`/`leading_zeros_u256` are branch-based, so it is the only unwound construct, and `unwind(12)` is ample.

- `..._matches_small_reference` — operands are exact multiples of `2^128`, so the result reduces to a native u128 division of the hi limbs and the loop must reproduce a known value bit-for-bit.
- `..._satisfies_division_identity` — symbolic low limb, checking `r < den` and `q * den + r == num`. That pair is unique, so the identity pins the result exactly, and it reaches the cross-limb borrows the first harness cannot express.

**Differentials in `cargo test`.** In `wide_math.rs`, against an independent base-2^32 schoolbook reference that shares no code with the divider: it multiplies and adds to rebuild the dividend, and never divides. Both directions — dividends constructed from a known `(q, r)` that must be recovered exactly, and arbitrary inputs checked against the identity — plus edge cases. Deterministic PRNG, `core` only, so it runs in the standard suite with no dependency to declare and nothing new to invoke.

**One thing I could not do.** I have no way to execute Kani here (Windows; Kani is Linux/macOS only), so I have not run those two harnesses — I am not going to claim a green result I did not see. To keep that from being a promise, `div_rem_u256_kani_harness_claims_hold_exhaustively` mirrors both harnesses over their exact input ranges under `cargo test`, and additionally asserts that the inputs do reach the loop, that `shift <= 3` so the unwind bound holds, and that both `cover!` conditions are satisfiable. If that passes and Kani still fails, the failure is mine and I will fix it.

`cargo fmt --check` and `cargo test --all-targets` are clean; the `E0433` is gone with the file that caused it. PTAL.

Ashutosh0x:
I installed Kani and ran the harnesses. Two corrections to my comment above, and the numbers.

**They are bound to production `div_rem_u256`.** Kani 0.67.0, x86_64-unknown-linux-gnu:

| harness | clean | loop guard `>=` → `>` | shift off-by-one |
|---|---|---|---|
| `..._matches_small_reference` | SUCCESSFUL (381s) | **FAILED** (463s) | SUCCESSFUL (402s) |
| `..._satisfies_division_identity` | SUCCESSFUL (542s) | **FAILED** (569s) | SUCCESSFUL (633s) |

Under the loop-guard mutation they fail on `assertion failed: q.lo() == hi_n / hi_d` and on `"remainder must be below the divisor"` respectively. Mutating the loop breaks the obligations, which is the thing my Z3 script could not do.

**Correction 1: the unwind bound is 260, not 12.** I wrote that the loop "runs at most 4 iterations and stays solver-tractable". The iteration count is right — the small nonzero `hi()` on both operands holds `shift <= 3` — but CBMC cannot use that. `leading_zeros()` lowers to a ctlz intrinsic it does not relate back to the assumed input range, so it unrolls to the full 256-bit worst case. Below 260 both harnesses fail on `div_rem_u256.unwind.0`, which is an unwinding assertion, not a property violation. Also worth knowing: `assert_eq!` on whole `U256` values pulls in a 32-byte `memcmp` whose own loop then needs unwinding, so the harnesses compare `.lo()`/`.hi()` limb-wise.

**Correction 2: the harnesses do not catch the shift off-by-one; the differentials do.** That is the last column above, and it is not a harness bug. Inside the small-`hi` input space, `den.shl(shift + 1)` has ~120 bits of headroom, so an over-large initial shift costs one no-op iteration and the result is still correct. The mutation only bites when that shift discards high bits, which needs full-width operands. The `cargo test` differentials catch it because they draw across the whole 256-bit range.

So the two layers are complementary rather than redundant, which is a better reason to keep both than the one I gave: Kani proves the loop bit-precisely over ~1.8M symbolic combinations the differentials only sample, and the differentials cover the operand widths the harnesses have to exclude to stay tractable.

**On cost.** ~6–10 minutes per harness against a budget the other arithmetic proofs in that file meet in seconds. If that is too expensive for `scripts/isolated_runner.sh`, dropping the two harnesses and keeping the differentials still catches all five mutations I tried, in under a second, and I would not argue against that.

Two notes for reproducing. `cargo kani` needs `--features fuzz` or it dies on `target reference_model_conformance ... requires the features: fuzz` before verifying anything; I used the same form as `scripts/isolated_runner.sh`: `cargo kani --tests --features fuzz --exact --harness <name>`. And `div_rem_u256_kani_harness_claims_hold_exhaustively` mirrors both harnesses over their exact input ranges, which turned out to be useful beyond portability — it predicted the shift off-by-one would be invisible in-range in under a second, which CBMC then took 402s to confirm.

PR #95, opened 2026-07-07
[codex] Recertify single trades on any account oracle lag

### LINKED ISSUE #93: single-trade recert applies the target-effective lag penalty only to the traded asset
Starting from this callsite https://github.com/aeyakovenko/percolator/blob/68986f7747ba8c42ee90336957ffc28833d6071d/src/v16.rs#L13737-L13742
, whenever we have single trades, thus recertify_after_fill = true, the two traders health is validated 

https://github.com/aeyakovenko/percolator/blob/68986f7747ba8c42ee90336957ffc28833d6071d/src/v16.rs#L13563-L13577

 while also taking into account the possible lag between effective_price and raw_oracle_target_price , caused by the clamped drift [towards oracle price](https://github.com/aeyakovenko/percolator-prog/blob/62c8abf06824ba70f4a38a8337486af4f35cf454/src/v16_program.rs#L4061-L4065) in these kinds of markets (we already mentioned this when tackling the bug fixed in b9a3783 ). Now, note that the branch here (https://github.com/aeyakovenko/percolator/blob/68986f7747ba8c42ee90336957ffc28833d6071d/src/v16.rs#L13451-L13459) only validates if the current asset is lagging: if some other asset in the portfolio is behind, no lag_loss_penalty (https://github.com/aeyakovenko/percolator/blob/68986f7747ba8c42ee90336957ffc28833d6071d/src/v16.rs#L10334-L10335) is applied to it. I believe the correct call is to have the health requirements take into account the lag for every asset, if any is present. oracle_epoch could be bumped to avoid another inner loop

aeyakovenko:
Fixes #93.

## Summary
- Make the single-fill recertification path fall back to full account health certification whenever any active account leg has raw-target/effective oracle lag, not only when the traded asset has lag.
- Add a regression where a trader holds a lagged asset 1 leg and attempts a single asset 0 trade that would pass only if the asset 1 lag penalty were ignored.
- Add focused Kani coverage proving the recert lag gate sees a lagged non-traded active asset.

## Root Cause
The single-trade fast path incrementally adjusted the certificate when the traded asset had no target/effective lag. That preserved a stale requirement for other active legs if a non-traded asset had acquired target/effective lag without bumping certificate epochs.

## Validation
- `cargo fmt --check`
- `git diff --check`
- `cargo test v16_single_trade_recert_applies_unrelated_asset_target_lag_penalty --test v16_spec_tests -- --nocapture`
- `cargo test --tests --features fuzz`
- `cargo kani --tests --features fuzz --exact --harness proof_v16_single_trade_recert_lag_gate_sees_nontraded_active_asset --output-format regular`

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -13447,8 +13447,7 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
         {
             return Err(V16Error::InvalidConfig);
         }
-        let asset = self.asset_state(asset_index)?;
-        if asset.raw_oracle_target_price != asset.effective_price {
+        if self.account_has_target_effective_lag(&account.as_view())? {
             let cert = self.compute_account_health_cert_with_price_override(
                 &account.as_view(),
                 false,
--- a/src/v16_kani_api.rs
+++ b/src/v16_kani_api.rs
@@ -783,6 +783,13 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
         Self::ensure_no_positive_credit_initial_margin(account)
     }
 
+    pub fn kani_account_has_target_effective_lag(
+        &self,
+        account: &PortfolioV16View<'_>,
+    ) -> V16Result<bool> {
+        self.account_has_target_effective_lag(account)
+    }
+
     pub fn kani_apply_trade_after_refresh_not_atomic(
         &mut self,
         long_account: &mut PortfolioV16ViewMut<'_>,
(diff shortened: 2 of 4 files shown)


aeyakovenko:
Superseded by #107, which fixes the root cause in O(1) by advancing oracle_epoch whenever the raw target changes. That globally invalidates prior health certificates and avoids this PR's active-leg scan. The stronger production-path Kani theorems and full engine suite passed in #107.

PR #96, opened 2026-07-09
[codex] Unblock terminal empty asset restart

aeyakovenko:
## Summary
- allow terminal empty Recovery/Retired assets to clear fully spent domain-budget residue
- allow flat source-backed PnL conversion in Recovery when no live risk exposure remains
- canonicalize matched consumed-only source/backing residue during terminal empty-asset cleanup

## Root Cause
A public liquidation/forfeit flow could leave an otherwise empty terminal asset with inert price reset fields, fully spent insurance-domain budget residue, and matched consumed-only source/backing counters. Those counters were not live withdrawable balances or open claims, but the empty-asset lifecycle gate treated them as active blockers, so restart/retire could be permanently blocked after normal user wind-down.

## Safety Shape
The cleanup still rejects any remaining unspent budget, positive source claim, fresh/valid/impaired backing, backing-provider fee earnings, insurance reservation, open interest, stale account, pending obligation, barrier, or loss state. It only clears terminal residue where budget == spent and source.spent == source.provider_receivable == bucket.consumed with no live backing amounts.

## Validation
- cargo test v16_terminal_spent_domain -- --nocapture
- cargo test v16_flat_source_claim_conversion_does_not_require_live_risk_cert -- --nocapture

Wrapper PR aeyakovenko/percolator-prog#191 pins to fb64ac40742803446c602c00d7dda38addb6af18 and proves the full public LiteSVM flow.

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -8637,6 +8637,37 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
         Ok((V16PodU128::default(), V16PodU128::default()))
     }
 
+    fn clear_terminal_source_backing_pair(
+        market_id: u64,
+        source: SourceCreditStateV16,
+        bucket: BackingBucketV16,
+    ) -> V16Result<(SourceCreditStateV16, BackingBucketV16)> {
+        if source.is_empty_amount_shape() && bucket.is_empty_amount_shape() {
+            return Ok((source, bucket));
+        }
+        if source.positive_claim_bound_num != 0
+            || source.exact_positive_claim_num != 0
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Superseded by #94, merged as `876fa9f`.

This broader branch is intentionally not merged. Its terminal source/backing cleanup cleared `provider_receivable_num` and `consumed_liened_backing_num`, but the frozen spec defines that pair as outstanding recoverable provider principal. The replacement fix clears only fully spent insurance-domain history and includes a production-path Kani exact-frame theorem proving nonzero provider receivables still reject without mutation.

PR #98, opened 2026-07-10
Skip Recovery legs in auto-crank selection

aeyakovenko:
## Finding

The self-classifying crank selected the first active portfolio leg without checking the selected asset's lifecycle. In a mixed portfolio where the first leg's asset was in `Recovery` and a later leg remained live, refresh and liquidation repeatedly selected the Recovery asset and returned `LockActive`.

## Fix

Ordinary refresh accrual and liquidation now select the first leg whose asset is `Active` or `DrainOnly`. Recovery legs remain account obligations and remain eligible for B-stale settlement, but are not selected for operations that reject Recovery assets. Recovery-only portfolios no longer advertise an impossible liquidation target.

## Broad proof

`proof_v16_recovery_legs_cannot_starve_dispatchable_auto_crank_work` symbolically partitions the full 16-leg account shape into disjoint Recovery, Active, DrainOnly, and inert masks. Across every ordering and multiplicity it proves:

- selected refresh/liquidation work is always Active or DrainOnly
- no Recovery leg can starve a later dispatchable leg
- first-match selection is deterministic
- Active and DrainOnly plans carry the exact selected asset
- Recovery-only portfolios expose no fake liquidation continuation

Result: 199 checks passed, 5/5 covers, 0.60s. Covers include Recovery before and after live work, Active and DrainOnly selection, and Recovery-only portfolios.

Mutation sensitivity: making Recovery lifecycle dispatchable fails the theorem at the lifecycle-equivalence assertion.

The production-state regression `v16_auto_crank_skips_recovery_first_leg_for_live_refresh` binds the symbolic kernel result to the persisted public auto-crank path.

## Validation

- `cargo test`: 50 unit + 9 backing fuzz + 2 resolved fuzz + 62 v16 spec tests
- `git diff --check`

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -11607,8 +11607,8 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
         );
         let ledger = account.header.close_progress.try_to_runtime()?;
 
-        let has_open_risk =
-            !active_bitmap_is_empty(account.header.active_bitmap.map(V16PodU64::get));
+        let (_, dispatchable_active_asset) = self.auto_crank_selected_assets(account)?;
+        let has_open_risk = dispatchable_active_asset.is_some();
         // A close ledger with residual_remaining==0 is already fully booked/covered
         // (e.g. insurance absorbed the loss); only OUTSTANDING residual is real,
         // actionable close work. The `active` flag can linger past that.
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #99, opened 2026-07-10
[HARDENING] Make prior-reset auto-crank obligations progress

aeyakovenko:
## Summary

- Make prior-reset obligations permissionlessly detachable without letting them starve a later live liquidation.
- Select refresh, liquidation, B-settlement, and reset-obligation assets from current engine state.
- Reuse that bounded leg selection during dispatch instead of scanning the portfolio twice.

## Public repro

Wrapper PR #201 constructs an account with an ADL-created prior-reset obligation before a separately liquidatable live leg. Before this change, every honest crank repeatedly selected the inert first leg and returned `EngineInvalidLeg`; the abandoned owner had to intervene before the live position could be liquidated.

The fixed route settles and detaches at most one prior-reset obligation per call, then exposes the later live liquidation to the next permissionless crank. Engine and wrapper calls remain one bounded action per transaction.

## CU integration correction

The first implementation scanned all bounded portfolio legs once while building the actionable summary and again before dispatch. Against current wrapper `main`, the existing issue-103 partial-liquidation crank rose from 295,076 CU to 328,570 CU and violated the 325,000-CU guard.

Commit `28a7338bd8c0dff7134ea0aa06404795bbb8557b` returns the already computed selections with the summary. The same public crank is now 318,594 CU and passes the existing guard without changing selection semantics.

## Validation

- Kani `proof_v16_prior_reset_cleanup_cannot_starve_live_liquidation`: passed
- Kani `proof_v16_recovery_legs_cannot_starve_dispatchable_auto_crank_work`: passed
- paired wrapper aggregate: 6 unit + 568 serialized LiteSVM/CU tests passed
- existing issue-103 crank: 318,594 CU, below 325,000
- `git diff --check`

Head: `28a7338bd8c0dff7134ea0aa06404795bbb8557b`.

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -11675,6 +11675,18 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
         &self,
         account: &PortfolioV16View<'_>,
     ) -> V16Result<ActionableSummaryV16> {
+        Ok(self
+            .build_actionable_summary_and_selected_assets(account)?
+            .0)
+    }
+
+    fn build_actionable_summary_and_selected_assets(
+        &self,
+        account: &PortfolioV16View<'_>,
+    ) -> V16Result<(
+        ActionableSummaryV16,
+        (Option<usize>, Option<usize>, Option<usize>, Option<usize>),
+    )> {
         let mode = decode_market_mode(self.header.mode)?;
         let live = mode == MarketModeV16::Live;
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Superseded by consolidated engine dependency #195. Every non-merge patch unique to #99 is patch-identical to a commit retained in #195 head 495a5590c97055bd71c6f94d849ff0298f243145.

PR #100, opened 2026-07-10
[REAL DoS] Keep partially ADL-reduced positions exit-live

aeyakovenko:
## Problem

A public partial liquidation can ADL a winning side so that an account retains two lots of stored basis while the market has only one lot of matched effective OI. A keeper or owner using the normal max-work amount then reaches `CounterUnderflow`. Reducing exactly one lot consumes all effective OI but leaves a nonzero leg in `Normal`, so every later reduction fails and no reset cleanup is actionable.

The LiteSVM reproduction uses only deposits, a public trade, authenticated mark updates, permissionless cranks, and owner `RebalanceReduce`.

## Fix

- Cap unilateral liquidation/rebalance work by matched effective OI.
- When a successful reduction exhausts a side's effective OI, enter the existing reset epoch.
- Classify and migrate legacy `Normal` zero-effective-OI residues so an unsigned auto-crank settles terminal K/F/B value and detaches them.
- Never start this migration while a loss barrier or pending obligation owns the side.

## Proof-driven validation

`proof_v16_unilateral_close_capacity_is_safe_progress_and_residue_complete` quantifies both sides, symbolic nonzero stored basis, matched effective OI, and arbitrary positive requested work over the production bounds. Through the exact production capacity, reduction, and exhausted-OI classifier kernels it proves:

- Unilateral work never exceeds either side's matched effective OI.
- Effective-OI subtraction cannot underflow.
- Reduction strictly decreases absolute basis and never flips the user's side.
- Partial requests, full account closes, and ADL terminal residues are all reachable.
- Any nonzero stored residue after matched OI reaches zero is classified as permissionlessly actionable.

Result: 109 checks, 5/5 covers, 1.48s.

Mutation sensitivity:

- Removing the matched-OI cap produces both a capacity assertion failure and checked-subtraction underflow.
- Disabling exhausted-OI residue classification fails the crank-actionability assertion with 4/5 covers still reachable.

Additional validation:

- Parent `a9baa08`: production regression fails with `CounterUnderflow`.
- Patched: 50 unit + 9 backing fuzz + 2 resolved fuzz + 65 spec tests pass.
- `cargo test --features fuzz`: 69 spec tests plus all differential/fuzz suites pass.
- Frozen `spec.md` unchanged.

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -15552,6 +15552,13 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
                 if self.has_pending_domain_loss_barrier(asset_index, leg.side)? {
                     return Ok(());
                 }
+                let asset = self.asset_state(asset_index)?;
+                if Self::leg_has_exhausted_effective_oi(asset, leg) {
+                    // A legacy Live state may resolve before its next auto-crank.
+                    // Establish the same terminal reset epoch here so resolved
+                    // teardown does not subtract stored basis from zero OI.
+                    self.begin_full_drain_reset_inner(asset_index, leg.side)?;
+                }
                 let (k_target, f_target) = self.kf_target_for_leg(asset_index, leg)?;
                 if k_target != leg.k_snap || f_target != leg.f_snap {
                     return Ok(());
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #101, opened 2026-07-10
[REAL DoS] Unstick abandoned ADL residues during Recovery

aeyakovenko:
## What changed

- Migrate a legacy nonzero-basis, zero-effective-OI leg into the normal reset epoch before recovery forfeit clears it.
- Add an engine regression for the exact ADL residue shape.

## Root cause

A partial ADL can leave stored basis larger than matched effective OI. A later matched abandoned-asset close can consume the last effective OI while leaving that stored residue. Recovery forfeit then called the normal clear path, which attempted to subtract the stored basis from zero OI and returned `CounterUnderflow` forever.

This is stacked on the partial-ADL fix in PR #100.

## Validation

- `cargo test`: 50 unit, 67 production spec, 9 backing fuzz, and 2 resolved-insolvency fuzz tests pass.
- The new engine regression fails on the parent behavior and passes with this migration.

## Formal proof

`proof_v16_full_drain_reset_then_prior_epoch_clear_is_total_and_exact` composes the same reset kernel used by production with `kernel_clear_leg`. Across symbolic sides, epochs, stored counts, position magnitudes, K/F/B targets, loss weights, and social-loss remainder/dust, it proves exact epoch migration, opposite-side framing, remainder quarantine, and underflow-free prior-epoch detach without subtracting current OI.

- Kani: 649/649 checks passed, 6/6 covers satisfied, 4.46s.
- Covers long/short, Normal/DrainOnly, nonzero remainder, multiple sibling residues, and nonzero K/F/B.
- Persisted Recovery regression additionally proves full-path detach, value neutrality, and post-state market/account validation.

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -1995,6 +1995,68 @@ impl V16Core {
         Ok(leg)
     }
 
+    /// PRODUCTION KERNEL: move an exhausted side behind a reset epoch without
+    /// touching the opposite side. Prior-epoch legs can then detach without
+    /// subtracting their stored basis from already-zero effective OI.
+    fn kernel_begin_full_drain_reset(
+        mut asset: AssetStateV16,
+        side: SideV16,
+    ) -> V16Result<AssetStateV16> {
+        match side {
+            SideV16::Long => {
+                if asset.mode_long == SideModeV16::ResetPending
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #102, opened 2026-07-10
[REAL DoS] Expire lapsed source backing during Live auto-crank

aeyakovenko:
## Finding

A fully public trade/partial-ADL sequence creates Fresh source backing for the winning account. Once that automatically-created bucket expires, a later adverse price move makes account refresh value the source claim. `validate_source_domain_ledger_current` returns `Stale`, but Live auto-crank had no expiry continuation. SVM rollback restores the expired bucket as `Fresh`, so every user or keeper crank repeats the same error and the position cannot make progress without privileged intervention or terminal resolution.

## Fix

The permissionless Refresh path scans the account's bounded source-domain roster and applies the existing canonical `expire_source_backing_bucket_not_atomic` transition to exactly the first lapsed Fresh bucket. It returns the explicit `SourceBackingExpired { domain }` progress outcome immediately. Repeated auto-cranks drain at most one domain per transaction, then the final call certifies the account.

The streaming selector now uses a small production scan-step kernel, preserving early exit and bounded CU while making its liveness semantics directly provable.

## Proof-driven validation

`proof_v16_live_source_backing_expiry_is_bounded_complete_and_isolated` quantifies compact roster length and every disjoint lapsed/future mask over the Kani-sized four-domain table. It composes the exact production scan-step kernel with the exact production expiry delta and proves:

- The first lapsed occupied domain is selected regardless of ordering.
- Clean rosters are no-ops and sparse tails terminate scanning.
- Every successful crank decreases the lapsed-domain rank by exactly one.
- All lapsed obligations are exhausted within the table bound.
- Future backing is never selected or mutated.
- Six clean/single/multiple/ordering/future/sparse-tail classes are reachable.

Result: 83 checks, 6/6 covers, 1.24s.

`proof_v16_counterparty_backing_expiry_reclassifies_principal_and_impairs_lien` independently quantifies fresh/liened backing and stock state through `MAX_VAULT_TVL`. It proves exact principal removal, exact lien impairment, no receivable/spent mutation, status classification, and pre/post vault-stock reconciliation.

Result: 172 checks, 4/4 covers, 3.30s.

Mutation sensitivity:

- Disabling lapsed-bucket selection fails first-match/rank progress.
- Selecting any Fresh bucket regardless of expiry fails future-domain isolation with 5/6 covers reachable.

## Additional validation

- Persisted two-domain production test proves one-domain-per-crank progress, unchanged vault/capital/insurance/earnings, aggregate backing release, risk-epoch advancement, recertification, and final shape validation.
- `cargo test --all-targets`: 136 passed (50 + 9 + 2 + 75).
- `cargo fmt --all`
- `git diff --check`
- Frozen `spec.md` unchanged.
- The stacked wrapper PR carries the public LiteSVM exploit, per-step CU, custody, and owner-exit assertions.

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -2527,6 +2527,28 @@ impl V16Core {
         Ok((bucket, source))
     }
 
+    /// PRODUCTION KERNEL: one step of the compact source-domain expiry scan.
+    /// The first sparse-tail entry terminates the scan; otherwise the first
+    /// occupied lapsed Fresh bucket is selected and terminates the scan.
+    fn kernel_lapsed_source_backing_scan_step(
+        selected: Option<usize>,
+        sparse_tail: bool,
+        occupied: bool,
+        domain: usize,
+        bucket_status: BackingBucketStatusV16,
+        expiry_slot: u64,
+        current_slot: u64,
+    ) -> (Option<usize>, bool) {
+        if selected.is_some() || sparse_tail {
+            return (selected, true);
+        }
+        if occupied && bucket_status == BackingBucketStatusV16::Fresh && expiry_slot <= current_slot
+        {
+            return (Some(domain), true);
+        }
+        (None, false)
+    }
+
     #[cfg(any(kani, feature = "fuzz"))]
     // Contract layer: lien release is the un-pledge relabel — valid liened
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #104, opened 2026-07-10
[REAL DoS] Bound resolved source-domain close work

aeyakovenko:
## What changed

- Split resolved source-domain teardown into bounded committed continuations.
- Release at most one source lien or expire at most one lapsed backing bucket per call.
- Settle one prepared source claim per call, realizing backed value into capital and receipting the remaining junior face.
- Keep final capital and receipt payout in the terminal close call.
- Route terminal payout arithmetic through a production kernel proved over the configured value range.

## Root cause

A public max-shape account can accumulate 28 historical source domains through ordinary trades across 14 assets. Resolved close previously swept all domains in one call. Both fresh and lapsed variants consumed the SVM ceiling of 1,400,000 CU, aborted, and rolled back forever.

## Broad proof-driven validation

`proof_v16_resolved_source_close_phase_strictly_decreases_global_rank` proves the production phase selector over every lien/lapsed/prepared state and an inductive tail representing all remaining domains. Every successful continuation decreases the exact global source-close rank by one, preserves the remaining obligations, reaches zero, and requires at most three steps per domain across the full 28-domain bound. Covers include max-domain lien release, expiry with a nonempty tail, prepared-claim progress, and the terminal zero-rank case.

`proof_v16_resolved_close_terminal_payout_conserves_vault_and_value_classes` quantifies account capital, resolved claim, vault, and `c_tot` through `MAX_VAULT_TVL`. It proves payout is exactly bounded by claim and custody, capital/receipt sources remain disjoint, and vault/value ledgers debit consistently across scarce and fully funded states.

The public fresh/lapsed parent scenarios fail at exactly 1,400,000 CU; fixed production continuations reduce the proved rank and converge. Production regressions bind the phase kernel to the public close route, so this is not only a model of intended behavior.

Additional checks:

- terminal payout theorem: 35 checks, 3/3 scarcity/funding covers;
- fresh and expired counterparty terminal-lien proofs;
- impaired-insurance terminal-lien proof;
- `cargo test --all-targets --quiet`: all engine tests pass;
- public wrapper PDD: https://github.com/aeyakovenko/percolator-prog/pull/205.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Source-checkable counts in this checkout:
 
 | Class | Count |
 | --- | ---: |
-| Plain Kani proofs in `tests/proofs_v16.rs` | 244 |
+| Plain Kani proofs in `tests/proofs_v16.rs` | 249 |
 | Plain Kani proofs in `tests/proofs_v16_arithmetic.rs` | 11 |
 | Plain Kani proofs in `src/v16_proofs.rs` | 38 |
 | Function-contract proofs in `src/v16_proofs.rs` | 51 |
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -14759,8 +14759,19 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
         if decode_bool(account.header.resolved_payout_receipt.present)? {
             return Ok(());
         }
-        self.initialize_resolved_payout_ledger_if_needed()?;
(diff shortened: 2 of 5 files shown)


aeyakovenko:
Superseded by #110, which consolidates this resolved source-domain CU fix with the active-leg and live max-source paths, rebased directly on master with the public wrapper repro and final CU measurements.

PR #106, opened 2026-07-10
[REAL DoS] Bound resolved active-leg teardown

aeyakovenko:
## What changed

- Resolve at most one ready active portfolio leg per `close_resolved_account_not_atomic` continuation.
- Inspect only the deterministic first active leg instead of preflighting and clearing the entire 14-leg portfolio.
- Return `ProgressOnly` immediately after every leg detach, including the final leg, so leg work never composes with source-domain settlement in the same transaction.
- Replace the proof-only Boolean `active_leg` rank with the exact active-leg count.
- Add a production-path engine test proving a two-leg account takes two strict progress steps before payout.

## Root cause

A public account can reach resolution with all 14 legs active and 28 historical source domains. The prior close path scanned and cleared every active leg, then entered source-domain settlement in the same call. Both direct close and resolved auto-crank exhausted the SVM hard ceiling, rolled back, and could never reduce state.

The old Boolean proof rank hid this frontier: it represented one active leg and fourteen active legs identically.

## Broad proof-driven validation

`proof_v16_resolved_leg_detach_exhausts_exact_rank_and_frames_other_obligations` quantifies every active-leg bitmap over the full 14-slot account, including sparse and maximum shapes. It repeatedly applies the production first-leg/rank kernels and proves:

- exactly one deterministic active leg is removed per continuation;
- active-leg rank decreases by one and reaches zero in exactly the initial popcount;
- the sequence is bounded by the account capacity;
- B-stale state, negative PnL, pending capital, receipt claims, and recovery requirements are framed unchanged while legs remain.

Five covers establish empty, singleton, maximum, sparse, and all-unrelated-obligations states. The production two-leg regression binds the theorem to the real zero-copy close path and requires two distinct committed steps before payout. The parent max-shape public scenario demonstrates the rollback DoS.

## Verification

- `cargo test --all-targets`
- `cargo kani --tests --features fuzz --harness proof_v16_resolved_leg_detach_exhausts_exact_rank_and_frames_other_obligations`
- `contract_check_build_resolved_close_rank`
- `contract_check_kernel_resolved_close_progress`
- Public max-shape LiteSVM TDD is in the stacked wrapper PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Source-checkable counts in this checkout:
 
 | Class | Count |
 | --- | ---: |
-| Plain Kani proofs in `tests/proofs_v16.rs` | 244 |
+| Plain Kani proofs in `tests/proofs_v16.rs` | 249 |
 | Plain Kani proofs in `tests/proofs_v16_arithmetic.rs` | 11 |
 | Plain Kani proofs in `src/v16_proofs.rs` | 38 |
 | Function-contract proofs in `src/v16_proofs.rs` | 51 |
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -276,6 +276,23 @@ fn active_bitmap_with_cleared(
 }
 
 #[inline]
+fn first_active_bitmap_slot(bitmap: V16ActiveBitmap) -> V16Result<Option<usize>> {
+    let mut word_index = 0usize;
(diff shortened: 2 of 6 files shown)


aeyakovenko:
Superseded by #110, which consolidates this resolved leg teardown fix with the source-domain and live max-source progress changes, rebased directly on master.

PR #110, opened 2026-07-10
[BLOCKER DoS] Bound max-source account progress

aeyakovenko:
## Finding

A fully public 14-asset portfolio with 28 occupied source domains could put both live liquidation progress and resolved terminal close at the 1,400,000 CU transaction ceiling. SVM rollback left the victim unchanged, and every available owner/keeper continuation repeated the same oversized work, so this was a persistent funds-locking state rather than a single expensive call.

During RED-to-GREEN verification, the first progress encoding also exposed a release blocker: global source domain IDs are not bounded by the 28-entry portfolio table. A `u32` bitmap rejected legitimate source claims on asset 16 and above even though markets publicly support more than 64 assets.

## Fix

- settle at most one stale K/F leg per permissionless continuation and reuse its health certificate
- linearize and fuse source-domain support/consumption work
- detach one resolved leg per call without full active-leg prescans
- realize at most one resolved source entry per call
- persist completion on each sparse source entry, so progress supports every configured global domain ID and follows entries through compaction
- restore domain-agnostic duplicate validation and validate per-entry progress as a resolved-only boolean
- keep layout discriminator 18 and prove each selected source-entry step strictly decreases its remaining rank

## Public regression and measurements

The paired wrapper PR constructs both states through public SBF instructions only. It covers the 14-leg/28-source max shape and a real backed winner on asset 16 (global domain 32), then requires finite permissionless progress and exact custody conservation.

Post-fix max CU:

- stale-leg settlement / refresh: 1,324,675 CU
- liquidation: 939,113 CU
- resolved close step: 1,163,297 CU
- resolved terminal calls for both max-shape accounts: 58

## Verification

- engine runtime: 141/141 passed
- terminal source-entry rank Kani harness: 0/8 failed, cover reached
- wrapper runtime against the exact rebuilt SBF: 6/6 + 567/567 passed
- wrapper instruction decoder Kani harness: 0/864 failed, cover reached

Wrapper public repro: aeyakovenko/percolator-prog#212

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Source-checkable counts in this checkout:
 
 | Class | Count |
 | --- | ---: |
-| Plain Kani proofs in `tests/proofs_v16.rs` | 248 |
+| Plain Kani proofs in `tests/proofs_v16.rs` | 249 |
 | Plain Kani proofs in `tests/proofs_v16_arithmetic.rs` | 11 |
 | Plain Kani proofs in `src/v16_proofs.rs` | 38 |
 | Function-contract proofs in `src/v16_proofs.rs` | 51 |
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Source-checkable counts in this checkout:
 
 | Class | Count |
 | --- | ---: |
-| Plain Kani proofs in `tests/proofs_v16.rs` | 244 |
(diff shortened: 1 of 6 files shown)


aeyakovenko:

PR #111, opened 2026-07-10
[REAL DoS] Unwind source liens when positive PnL reverses

aeyakovenko:
## What changed

- Unwind the minimum counterparty/insurance source lien required when positive PnL face is burned.
- Preserve lien support still valid for the remaining claim and prorate capital-at-risk fee history.
- Route production mutation through one face-partition/backing-release plan so the proof and deployed path share the kernel.
- Keep terminal insurance-release semantics while exposing a production-safe internal Live release path.

## Root cause

A normal risk-increasing trade could use backed positive PnL for initial margin and create a source lien. If the mark later reversed, K/F settlement had to reduce that positive PnL, but Live-mode claim burning only consumed unliened face. The lien-release helper first attempted the same blocked settlement, so crank, trade reduction, unilateral reduction, and liquidation all returned `LockActive` indefinitely.

The fix treats disappearing claim face as an unwind, not a backing consumption: realizable unliened support still offsets the loss, while still-liened provider principal is returned before the associated face is burned.

## PDD validation

The proofs are symbolic invariants over the production planning kernels, not fixed-case unit checks:

- `proof_v16_source_lien_face_burn_partition_is_total_disjoint_and_strict_progress`: full-width `u128` source faces and burn; exact disjoint partition, no omission, and strict aggregate face-rank decrease. 32 checks, 4/4 cover goals, 1.88s.
- `proof_v16_source_lien_face_burn_plan_is_total_minimal_and_source_isolated`: both source classes, partial/full/cross-source reversal, minimum required release, independent post-state bounds, and rank-zero completion. 93 checks, 6/6 cover goals, 15.16s.
- `proof_v16_source_lien_face_burn_releases_only_required_backing`: fractional face/backing rounding. 41 checks, 3/3 cover goals, 12.91s.
- `proof_v16_source_lien_fee_proration_tracks_remaining_backing`: fee-history conservation under release. 406 checks, 3/3 cover goals, 8.46s.
- Mutation check: restoring the pre-fix `LockActive` behavior fails the broad composition theorem.
- Mutation check: assigning the full burn to insurance overlaps sources and fails the full-width partition theorem.
- `cargo test --all-targets`: 130 passed.
- `spec.md` remains unchanged.
- Wrapper lifecycle coverage: https://github.com/aeyakovenko/percolator-prog/pull/213

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -717,6 +717,27 @@ impl V16Core {
     }
 
     #[inline]
+    fn source_lien_backing_release_for_face_burn(
+        face_locked_num: u128,
+        backing_reserved_num: u128,
+        face_burn_num: u128,
+    ) -> V16Result<u128> {
+        if face_burn_num > face_locked_num || backing_reserved_num % BOUND_SCALE != 0 {
+            return Err(V16Error::InvalidLeg);
+        }
+        let effective_reserved = backing_reserved_num / BOUND_SCALE;
+        if effective_reserved > Self::amount_from_bound_num(face_locked_num)? {
+            return Err(V16Error::InvalidLeg);
+        }
+        let remaining_face_num = face_locked_num
+            .checked_sub(face_burn_num)
+            .ok_or(V16Error::CounterUnderflow)?;
+        let max_remaining_effective = Self::amount_from_bound_num(remaining_face_num)?;
+        let effective_release = effective_reserved.saturating_sub(max_remaining_effective);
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Superseded by consolidated engine dependency #195. Every non-merge patch unique to #111 is patch-identical to a commit retained in #195 head 495a5590c97055bd71c6f94d849ff0298f243145.

PR #112, opened 2026-07-11
[REAL DoS] Reconcile expired source liens during live refresh

aeyakovenko:
## What changed

- Treat a lapsed source-backing domain as actionable even when price and funding epochs are unchanged.
- Reconcile exactly one live account lien from an expired source domain per auto-crank, relabeling its claim as impaired before any source-credit health read.
- Route both stale classification and mutation through one production scan-step/selection kernel.
- Cap backing-utilization rent at the bucket expiry slot.
- Return bounded auto-crank work so a max-source account progresses through a finite, rank-decreasing sequence instead of an atomic full-domain mutation.
- Add production lifecycle, mixed-fee impairment, bounded-progress, and Kani regressions.

## Root cause

A portfolio could create a valid source-credit IM lien through ordinary trades while its backing bucket was fresh. Once the configured bucket expiry arrived, source-credit health reads returned `Stale`, but no public route reconciled the account lien. Crank, bilateral reduction, unilateral reduction, and PnL conversion could all fail, and depositing more capital did not repair the state.

This contradicted the expiry rule that aggregate backing expires before a credit-rate read and account liens become impaired through bounded permissionless progress.

## Impact

A backing authority could publish short-lived backing, allow users to create real liens, and then leave a live position unable to progress after expiry. The corrected route needs at most one successful crank per live source-domain obligation and does not require a max-domain atomic mutation.

## PDD validation

The proofs are symbolic invariants over production kernels, not fixed-case unit checks:

- `proof_v16_live_source_backing_reconcile_is_bounded_progress`: every ordering of lapsed backing, lapsed account liens, already-impaired account liens, future liens, and foreign-account impaired backing in the four-domain Kani roster. It proves first-action completeness, exactly one rank decrease per successful call, completion within the roster bound, future-lien isolation, and foreign-account isolation. 295 checks, 8/8 cover goals, 8.95s.
- `proof_v16_expired_counterparty_lien_impairment_is_exact_relabel`: symbolic face, effective backing, mixed insurance support, prior impairment, and fee history; exact relabel and conservation. 245 checks, 3/3 cover goals, 21.14s.
- Mutation check: removing already-impaired account-lien selection fails the bounded-progress theorem.
- Mutation check: classifying future Fresh buckets as expired fails the bounded-progress theorem.
- `cargo test --all-targets`: 140 passed.
- `spec.md` remains unchanged.
- Wrapper max-shape lifecycle coverage: https://github.com/aeyakovenko/percolator-prog/pull/214

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -716,6 +716,65 @@ impl V16Core {
         Ok((required_face_num, required_backing_num))
     }
 
+    fn prepare_account_counterparty_lien_impairment(
+        mut source: PortfolioSourceDomainV16Account,
+    ) -> V16Result<(PortfolioSourceDomainV16Account, u128)> {
+        let face = source.source_claim_counterparty_liened_num.get();
+        let backing_num = source.source_lien_counterparty_backing_num.get();
+        if face == 0 && backing_num == 0 {
+            return Ok((source, 0));
+        }
+        if face == 0 || backing_num == 0 || backing_num % BOUND_SCALE != 0 {
+            return Err(V16Error::InvalidLeg);
+        }
+        let effective = backing_num / BOUND_SCALE;
+        if effective == 0 {
+            return Err(V16Error::InvalidLeg);
+        }
+
+        source.source_claim_counterparty_liened_num = V16PodU128::new(0);
+        source.source_claim_liened_num = V16PodU128::new(
+            source
+                .source_claim_liened_num
+                .get()
+                .checked_sub(face)
+                .ok_or(V16Error::CounterUnderflow)?,
+        );
+        source.source_claim_impaired_num = V16PodU128::new(
+            source
+                .source_claim_impaired_num
+                .get()
+                .checked_add(face)
+                .ok_or(V16Error::CounterOverflow)?,
+        );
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Independent production-SBF confirmation of the DoS fixed here:

On `percolator-prog@da64be6` / engine `143e68c`, a public trade created a 443-atom source lien backed by a bucket expiring at slot 105. At slot 105 the bucket remained `Fresh`, the lien remained fully live, and the certificate reported zero deficit even though equity without that expired support was 647 against a 1,043 maintenance requirement.

Three additional permissionless cranks returned success with byte-identical market/portfolio state. Owner `RebalanceReduce` and a bilateral reducing `TradeNoCpi` both returned `EngineStale`. Depositing 1,000 more atoms (capital 1,600, independently above maintenance) still did not restore owner reduction.

This confirms **REAL DoS** reachability and persistence through the public wrapper, including the successful-no-op crank branch on the exact current pin. The bounded one-domain reconciliation in this PR matches the missing transition. The PR currently reports a conflicting stacked base and will need rebasing before it can be merged.

aeyakovenko:

PR #113, opened 2026-07-11
Bound max-source lien allocation CU

aeyakovenko:
Public LiteSVM construction can retain 20 source-claim domains. After one source lien is allocated, a one-position risk increase that needs a second lien exhausts the 1.4M SVM budget.

The per-domain allocator redundantly ran full account/market validation before and after each allocation, even though the allocating loop validates once and the trade runs its committed-state audit. Convert the inner helper to an explicitly named core and rely on those existing boundary validations.

The paired wrapper TDD reproduces the public lifecycle and measures the fixed second-lien trade at 898,270 CU. Kani does not model SBF compute units, so that LiteSVM regression is the direct DoS witness.

## Proof-driven verification

The production allocator and persisted account-source aggregate builder are shared directly by runtime validation and the proofs.

- `proof_v16_source_lien_allocation_plan_is_complete_bounded_and_ordered` symbolically composes four independent source domains with arbitrary claims, backing, rates, and request size. It proves exact greedy allocation, per-source claim/backing caps, zero allocation after completion, sufficient/insufficient aggregate behavior, and reachability of every source. Result: 208 checks, 5/5 covers, 37.49s.
- `proof_v16_source_lien_entry_relabel_preserves_persisted_aggregate_invariant` applies symbolic mixed counterparty/insurance liens through production mutation functions, then validates the persisted aggregate invariant used by full account validation. It proves partition totals, exact effective/backing scaling, claim-bound preservation, fee-cursor behavior, and impaired-state noninterference. Result: 243 checks, 4/4 covers, 7.55s.
- Existing production-path lien creation and scaled-unit proofs also pass: 121 checks in 1.05s and 422 checks with 3/3 covers in 194.77s.
- Mutation checks are effective: removing the backing cap fails the allocation theorem; omitting the counterparty backing aggregate update fails persisted validation.
- `cargo test --all-targets`: 139 passed, 0 failed.
- `cargo fmt --all -- --check` and `git diff --check`: pass.
- `spec.md` is unchanged.

Stacked on the max-source refresh branch because that branch makes the 20-domain public setup reachable.

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -11607,8 +11607,8 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
         );
         let ledger = account.header.close_progress.try_to_runtime()?;
 
-        let has_open_risk =
-            !active_bitmap_is_empty(account.header.active_bitmap.map(V16PodU64::get));
+        let (_, dispatchable_active_asset) = self.auto_crank_selected_assets(account)?;
+        let has_open_risk = dispatchable_active_asset.is_some();
         // A close ledger with residual_remaining==0 is already fully booked/covered
         // (e.g. insurance absorbed the loss); only OUTSTANDING residual is real,
         // actionable close work. The `active` flag can linger past that.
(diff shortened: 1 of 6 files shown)


aeyakovenko:

PR #114, opened 2026-07-11
[REAL DoS] Commit recovery-required auto-crank progress

aeyakovenko:
## What changed

- Convert a liquidation `RecoveryRequired` into successful crank progress only when that same call left a complete Recovery mode + reason marker.
- Preserve every other engine error unchanged, including bare `RecoveryRequired`, incomplete Recovery state, and unrelated errors while Recovery is set.
- Let the next auto-crank perform the value-neutral Recovery-to-Resolved transition.
- Avoid duplicate liquidation sizing, leg scans, and residual-capacity work on the hot path.

## Root cause

A liquidation could declare Recovery and then return `RecoveryRequired`. The wrapper propagated the error, so SVM rollback erased the declaration and permissionless keepers could repeat the same failing transaction forever.

## PDD evidence

- Red, engine parent `143e68c`: the public LiteSVM crank fails `Custom(23)` after 95,157 CU and Recovery rolls back.
- Kani `proof_v16_liquidation_error_commits_only_fully_declared_recovery`: exhaustive over all 12 engine errors, all 3 market modes, all 8 reasons plus missing reason; 296 checks, 4/4 covers, 0.126s.
- Engine integration drives a real two-leg uncovered-loss liquidation through the reactive Recovery path, checks value/account neutrality, then reaches Resolved on the next crank.
- Wrapper PR #190 runs the same behavior through the public instruction and SVM rollback boundary.

## Validation

- `cargo kani --tests --features fuzz --harness proof_v16_liquidation_error_commits_only_fully_declared_recovery`: passed
- Wrapper exact recovery regression: passed
- Wrapper issue-103 partial-liquidation CU guard (325k): passed
- Wrapper full suite: 5 unit + 564 LiteSVM tests passed
- `git diff --check`

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -172,10 +172,10 @@ pub fn v16_domain_pair_for_asset_index(asset_index: usize) -> V16Result<(usize, 
 /// price. `RefreshAccount { asset_index: Some(_) }` and EVERY other plan are
 /// dispatchable from committed on-chain state alone — `SettleBChunk` ignores
 /// price, `Liquidate` reads the current health cert, `DeclareRecovery` /
-/// `CloseResolved` / `NoAction` take no price — so a keeper holding no fresh
-/// observation can still drive the account forward (no liveness stall). A wrapper
-/// may call this to decide whether it must source an oracle observation before
-/// cranking.
+/// `FinalizeRecovery` / `CloseResolved` / `NoAction` take no price — so a keeper
+/// holding no fresh observation can still drive the account forward (no liveness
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #114 (bd77e34739087a2f442792374e228af0ddd5aeba) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #115, opened 2026-07-11
[REAL DoS] Allow resolved close after ADL scaling

aeyakovenko:
## What changed

- Add an ADL-aware terminal clear kernel for resolved positions.
- Remove `min(abs(stored_basis_q), remaining_oi_eff_side)` from current-epoch effective OI while removing the full stored-position count and loss weight.
- Detach prior-reset records without debiting already-reset side aggregates again.
- Keep `spec.md` frozen.

## Root cause and impact

Quantity ADL scales aggregate effective OI while preserving each winner stored basis as a settlement record. Resolved close reused the live clear kernel, which required effective OI to cover the full stored basis. After a normal partial liquidation, `CloseResolved` could therefore return `CounterUnderflow` forever and permanently strand a winner leg and payout.

Wrapper PR #216 reproduces this using only deposits, trade, authenticated mark movement, permissionless liquidation, resolution, and public close. The parent returns `Custom(25)` at about 128k CU; the fixed lifecycle fully reconciles both users and custody.

## PDD evidence

- `proof_v16_resolved_leg_clear_closes_every_adl_gap_without_cross_side_mutation`: symbolic strict ADL gaps on both sides plus prior-reset records; 157 checks, 3/3 covers, 2.33s.
- `contract_check_kernel_clear_resolved_leg`: arbitrary-input success-frame contract; 1,290 checks, 9.9s.
- Engine integration closes an ADL-scaled resolved account with stored basis above effective OI and reconciles OI, count, loss weight, vault, and `c_tot`.
- Public LiteSVM red/green and custody theorem in wrapper PR #216.

## Validation

- Both Kani harnesses above: passed under 300s
- Wrapper exact lifecycle regression: passed
- Wrapper full suite: 5 unit + 564 LiteSVM tests passed
- `git diff --check`

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1495,6 +1495,8 @@ For every recovered account, the engine MUST compute `recovery_value_transfer_bo
 
 Recovery preserves and reconciles `SourceCreditState`, insurance-credit reservations, lien buckets including impaired buckets, liens, close ledgers, pending obligations, B/ADL/support progress, and payout ledgers. It cannot erase ledgered progress and recompute gross loss. Recovery direct positive payout must use resolved receipts or pay no positive junior value.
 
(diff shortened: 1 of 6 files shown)


aeyakovenko:

PR #116, opened 2026-07-11
[REAL DoS] Keep social-loss dust from blocking owner exit

aeyakovenko:
## What changed

- Fold canonical per-leg social-loss remainders modulo `SOCIAL_LOSS_DEN` instead of rejecting when side dust carries one atom.
- Rebook a carried atom into the remaining side B-index, advancing `risk_epoch`, or conservatively charge the exiting account when no recipient weight exists.
- Enforce canonical social-loss remainder and dust bounds in the production audit.

## PDD

Wrapper PR: https://github.com/aeyakovenko/percolator-prog/pull/217

The public LiteSVM regression creates four matched positions, realizes a two-atom bankruptcy, and gives two exiting winners exactly half an atom of B remainder each. Before the fix, the second ordinary owner exit permanently returned `Custom(23)` at about 114k CU because SVM rollback restored the same carry state.

The engine proof now executes the production `kernel_clear_leg` over symbolic side, canonical leg remainder, and canonical current dust. It requires successful closure, zero OI/count/weight, and exact modulo dust, with explicit carry and no-carry cover properties.

Mutation check: restoring the old `RecoveryRequired` carry branch makes the closure proof fail at the production call and loses the carry cover. This prevents the former weak `Err(_) => true` behavior from proving the vulnerable implementation.

## Validation

- Kani `proof_clear_leg_canonical_dust_never_blocks_exit`: 177 checks, 2/2 covers, passed in about 1.5s
- Kani `contract_check_kernel_fold_social_loss_dust`: 227 checks, 2/2 covers, passed
- Kani `contract_check_kernel_clear_leg`: 1,634 checks, passed
- wrapper stack: 5 unit + 564 LiteSVM/SBF tests passed
- `git diff --check`

Engine commit: `4beb6ce141bb28656e7cc38b82a3d5c54e1884e4`

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -1928,11 +1928,48 @@ impl V16Core {
         Ok(leg)
     }
 
+    /// Fold one exiting leg's fractional social-loss remainder into the
+    /// side-level dust bucket. Both inputs are canonical fractions, so the sum
+    /// carries at most one whole quote atom.
+    #[cfg_attr(all(kani, feature = "contracts"), kani::requires(
+        current_dust < SOCIAL_LOSS_DEN && leg_remainder < SOCIAL_LOSS_DEN
+    ))]
+    #[cfg_attr(all(kani, feature = "contracts"), kani::ensures(|result: &V16Result<(u128, u128)>| match result {
+        Ok((dust, carry)) => {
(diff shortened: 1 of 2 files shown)


aeyakovenko:

PR #117, opened 2026-07-11
[REAL DoS] Quarantine reset carry without blocking progress

aeyakovenko:
## Summary

- Fold a canonical side-wide B remainder into side dust during full-drain reset.
- Record a whole carry as durable `explicit_unallocated_loss` instead of returning `RecoveryRequired` forever.
- Treat canonical dust and explicit loss as audit-only once positions, B index, funded stock, and barriers are empty.

## PDD

Wrapper PR: https://github.com/aeyakovenko/percolator-prog/pull/218

The public regression creates two normal bankruptcy cycles. At the second zero-OI reset, canonical B remainder plus canonical side dust crosses one atom. Before the fix, each liquidation returned `Custom(23)` at about 188k CU and SVM rollback restored the same underwater account.

The production-kernel contract now requires `kernel_quarantine_social_loss_remainder` to succeed for every canonical dust/remainder pair with representable explicit-loss headroom. It proves exact modulo dust and exact carry attribution, with separate carry and no-carry cover properties.

Mutation check: restoring the old reject-on-carry implementation makes the contract fail and removes the carry cover. The prior `Err(_) => true` proof weakness is removed.

## Validation

- Kani `contract_check_kernel_quarantine_social_loss_remainder`: 355 checks, 2/2 covers, passed in about 1.1s
- wrapper stack: 5 unit + 565 LiteSVM/SBF tests passed
- `git diff --check`

## Stack

Base: engine PR #116
Engine commit: `552e9ac76a2288fb5a91c40e1cce7e952a6ab4a5`

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -1960,6 +1960,36 @@ impl V16Core {
         }
     }
 
+    /// Move a side-wide B-division remainder into the durable dust/audit
+    /// buckets before the side's weight basis is reset. The two canonical
+    /// fractions can produce at most one whole quote atom; that atom is an
+    /// explicit unallocated loss, not account health or payout capacity.
+    #[cfg_attr(all(kani, feature = "contracts"), kani::requires(
+        current_dust < SOCIAL_LOSS_DEN && social_remainder < SOCIAL_LOSS_DEN
+    ))]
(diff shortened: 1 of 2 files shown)


aeyakovenko:

PR #118, opened 2026-07-11
[REAL DoS] Reset both zero-OI sides after unilateral close

aeyakovenko:
## Summary

- Detect when the liquidated side itself reaches zero effective OI while stored legs remain.
- Enter the same bounded `ResetPending` old-epoch route already used for the matching side.
- Prove both orientations through the production zero-copy mutable view, not only through the scalar predicate.

## Public red case

Wrapper PDD: https://github.com/aeyakovenko/percolator-prog/pull/219

After the two-bankruptcy lifecycle from PR #218, both effective-OI sides are zero but one short account remains stored. On the parent engine, only the matching long side enters `ResetPending`; owner `ForfeitRecoveryLeg` returns `Custom(21)` / `LockActive`, ordinary rebalance has no matching OI, and the user cannot become flat or withdraw.

With this PR, the liquidated side also enters `ResetPending`; the owner dead-leg exit, reset finalization, and complete SPL withdrawal all succeed within the custody CU limit.

## PDD strength

The original 105-check contract only proved a Boolean helper and could not catch deletion of its production call. The added composition theorem constructs a one-slot market, chooses long or short symbolically, invokes the actual `reduce_matching_open_interest_for_unilateral_close` view path, and proves both OI values are zero and both modes are `ResetPending`.

Mutation check: disabling only the closed-side production transition makes the composition proof fail one assertion and leaves both orientation covers unreachable.

## Validation

- composition theorem: 4,794 checks, 2/2 orientation covers, passed in 143s
- scalar contract: 105 checks, 2/2 reset/non-reset covers, passed
- mutation run: failed as expected
- wrapper stack: 5 unit + 565 LiteSVM/SBF tests passed
- `git diff --check`

## Stack

Base: engine PR #117
Engine commit: `f97194677901e38bc8b2b87a67ac878c74dbfa99`

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -1992,6 +1992,28 @@ impl V16Core {
         Ok((dust, explicit))
     }
 
+    /// A unilateral close mutates both effective-OI sides. If the liquidated
+    /// side itself reaches zero while other stored legs remain, those legs
+    /// need the same old-epoch reset route as the matching side; otherwise
+    /// they remain live-looking but cannot reduce or use the dead-leg route.
+    #[cfg_attr(all(kani, feature = "contracts"), kani::ensures(|result: &bool| {
+        *result == (effective_oi == 0
+            && stored_position_count != 0
+            && pending_obligation_count == 0
+            && mode != SideModeV16::ResetPending)
+    }))]
+    pub(crate) fn kernel_side_needs_full_drain_reset(
+        effective_oi: u128,
+        stored_position_count: u64,
+        pending_obligation_count: u64,
+        mode: SideModeV16,
+    ) -> bool {
+        effective_oi == 0
+            && stored_position_count != 0
+            && pending_obligation_count == 0
+            && mode != SideModeV16::ResetPending
+    }
+
     /// PRODUCTION KERNEL: the clear-leg asset transform — decrement the
     /// side's stored-position count (and pending-obligation count for a
     /// zero-basis obligation leg), and unless the leg predates a side reset,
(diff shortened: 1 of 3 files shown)


aeyakovenko:
A distinct public bilateral-close variant remains reachable after this fix.

Reproduction: `aeyakovenko/percolator-meta@9c75c9d` on branch `agent/real-dos-bilateral-zero-oi-terminal`, test `e2e_social_loss_then_bilateral_close_has_a_bounded_terminal_exit`.

Public sequence:
1. A solvent long opens 20 against a solvent short, then 1 against a small funded short.
2. An honest mark bankrupts the small short and social-loss settlement reduces long effective OI from 21 to 20 while the solvent winner retains basis 21.
3. The solvent pair legitimately closes the remaining matched 20 bilaterally.
4. Both effective OI counters become zero, but `stored_pos_count_long == 1`; the winner retains normal-epoch basis 1 and `mode_long == Normal`.

The isolated LiteSVM probe exhausts the bounded continuations: auto-crank makes no rank progress; `FinalizeResetSide` is a successful byte-no-op; crossing trade is H-locked; `RebalanceReduce` rolls back with counter underflow; drain-only does not change that; recovery `ForfeitRecoveryLeg` underflows; force-close has no opposite stored leg; and resolved `CloseResolved` underflows. The winner still has all 100,000 deposited atoms and remains the sole `resolved_payout_blocker_count`.

I also ran the same probe against the real SBF built from the program counterpart #219 (`b0975e5`, SHA-256 `d910e21249d04310bd01ea545544d6dced561b2480e54aace2b8cf192192df0e`); it fails identically. The current fix runs only from the unilateral-close path, while this transition reaches zero OI in a later bilateral trade. Please extend the invariant repair to that transition and add this public regression.

aeyakovenko:

PR #119, opened 2026-07-12
[HARDENING] Keep fragmented risk reductions live below IM

aeyakovenko:
## What changed

- Gate post-trade initial margin only when a fill increases risk or either post-trade account is liquidatable.
- Keep the locked-lane no-positive-credit IM gate scoped to that same admission/bankruptcy boundary.
- Leave the frozen `spec.md` unchanged.

## Root cause and impact

The trade preflight already proves that a matched risk-reducing fill does not increase either counterparty's absolute position. The final trade stage nevertheless required both accounts to cover initial margin after every fill.

That creates a permissionless recovery deadlock when one maintenance-healthy account has a large position and aggregate opposite OI is fragmented across smaller portfolios. Each available force-close reduces both sides, but each partial fill reverts until one counterparty is large enough to restore the large account's IM in a single call. No such counterparty has to exist.

Risk-increasing fills and liquidatable accounts retain the existing initial-margin and locked-lane checks, so bilateral trading cannot bypass bankruptcy attribution. Two-sided reductions where both post-trade accounts remain maintenance-healthy can decrease position rank below IM, preserving user and delayed recovery exit liveness.

## Proof-driven verification

`proof_v16_trade_final_margin_policy_preserves_exit_liveness_and_bankruptcy_gates` is a symbolic theorem over the production final-margin policy, not a fixed trade fixture. It quantifies both accounts' capital, positive/negative PnL, fee debt, certified equity, IM requirement, liquidation deficit, certificate validity, risk direction, and locked-lane state. It proves the full policy partition:

- maintenance-healthy bilateral reductions do not require same-fill IM recovery;
- every risk increase still requires both certified IM checks;
- any liquidatable post-state loses the reduction exemption;
- locked risk cannot satisfy IM with positive PnL credit;
- fully principal-funded locked risk remains admissible.

All five policy classes have explicit reachability covers. Restoring the parent unconditional final-IM gate falsifies the reduction-liveness assertion; weakening either bankruptcy gate falsifies the corresponding safety assertion. The persisted production regression and public wrapper LiteSVM scenario bind the theorem to the full trade path.

## Validation

- `cargo fmt --check`
- `cargo test --all-targets`
- `cargo kani --tests --features fuzz --harness proof_v16_trade_final_margin_policy_preserves_exit_liveness_and_bankruptcy_gates`
- Focused regression: `v16_two_sided_reduction_does_not_require_same_fill_im_recovery`
- Existing bankrupt/off-mark recycle regressions remain rejecting

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Source-checkable counts in this checkout:
 
 | Class | Count |
 | --- | ---: |
-| Plain Kani proofs in `tests/proofs_v16.rs` | 244 |
+| Plain Kani proofs in `tests/proofs_v16.rs` | 245 |
 | Plain Kani proofs in `tests/proofs_v16_arithmetic.rs` | 11 |
 | Plain Kani proofs in `src/v16_proofs.rs` | 38 |
 | Function-contract proofs in `src/v16_proofs.rs` | 51 |
--- a/spec.md
+++ b/spec.md
@@ -1465,6 +1465,7 @@ Trades require:
 - current B/K/F settlement for touched legs;
 - side-mode gating;
 - OI/position bounds;
+- post-trade initial margin for risk-increasing fills; a matched fill that strictly reduces both counterparties' absolute positions MUST remain admissible below initial margin so fragmented exits can make bounded progress;
 - candidate-slippage neutralization;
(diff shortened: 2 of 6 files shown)


aeyakovenko:

PR #120, opened 2026-07-12
[REAL DoS] Scope bankruptcy locks to affected assets

aeyakovenko:
## Bug

A single-asset bankruptcy set a global H-lock that permanently blocked favorable value exits for accounts referencing only unrelated assets. The first PR revision inferred the affected asset from mutable state such as `DrainOnly`, B state, insurance spend, and barriers. That inference was unsound in both directions: ordinary lifecycle state could fabricate attribution, while cleanup could erase the original attribution.

## Fix

- persist a canonical bankruptcy bit in each zero-copy asset slot
- maintain an exact header count so marking and asset-reuse teardown remain O(1)
- preserve unattributed events as globally fail-closed
- treat active legs, source claims, and the active close ledger as independent reference carriers
- clear only the reused asset bit; clear the global attributed lock only when the exact count reaches zero
- leave the frozen `spec.md` unchanged

## Proof-driven verification

The proofs quantify symbolic lock sets, wire states, account reference combinations, and lifecycle transitions over production helpers. They are not fixed-output unit assertions.

- `proof_v16_bankruptcy_hlock_attribution_is_asset_local_and_fail_closed`: 3248 checks, 6/6 covers, 148.9s
- `proof_v16_bankruptcy_asset_lock_is_durable_not_inferred_from_transient_state`: 406 checks, 2/2 covers, 1.0s
- `proof_v16_mark_attributed_bankruptcy_is_exact_durable_and_idempotent`: 4587 checks, 3/3 covers, 59.8s
- `proof_v16_bankruptcy_asset_reuse_clears_exactly_one_attribution`: 1249 checks, 3/3 covers, 3.5s
- adjacent fail-closed H-lock proofs: 3868 and 3217 checks, all covers reached

Mutation checks demonstrate sensitivity: restoring transient-state inference fails the durability theorem, and omitting close-ledger references fails the isolation theorem.

Wrapper reproduction and CU coverage remain in [percolator-prog #229](https://github.com/aeyakovenko/percolator-prog/pull/229).


aeyakovenko:

PR #124, opened 2026-07-15
Keep funded flat assets reusable

aeyakovenko:
## Impact

A normal funded asset can become permanently unreusable after its trading lifecycle is complete. Balanced OI advances the funding indices; after both users settle, close, and withdraw all source backing, the asset has no positions, stale accounts, obligations, loss state, source claims, backing, reservations, K, B, or value claimant. Historical current and epoch-start F values remain, and the parent empty-lifecycle guard rejects retirement/restart forever.

This is a partial public DoS against finite market-slot capacity. The trigger uses ordinary matched trading and funding accrual, not malformed or directly injected state.

## Fix

Treat current and epoch-start F indices as inert settlement anchors only after every claimant-bearing and value-bearing field satisfies the existing empty-state requirements. Retirement preserves the history; restart changes `market_id` and canonicalizes the slot.

All real blockers remain enforced: side mode, ADL epoch shape, K/B, OI, position/stale/obligation counts, loss weights, social-loss remainder/dust, explicit loss, barriers, source credit, backing, reservations, and spent insurance.

## Proof-driven evidence

The source extracts the exact O(1) asset-local empty-lifecycle predicate without changing behavior, making the semantic boundary directly provable.

- `proof_v16_empty_lifecycle_classifier_is_f_history_independent` quantifies all four F lanes as arbitrary `i128` values (512 symbolic bits) across both valid ADL epoch shapes. It proves classification is identical to zero-F state. Result: 39 checks, 2/2 full-width covers, 0.113s.
- `proof_v16_empty_lifecycle_classifier_rejects_every_asset_claimant_and_loss_field` exhausts all 14 asset-local blocker classes on both sides while nonzero/extreme F history coexists. Result: 14 checks, 3/3 covers, 0.063s.
- `proof_v16_retire_empty_asset_is_value_neutral_and_epoch_scoped` calls the real zero-copy retirement path with all four F histories present. It proves success, history preservation, exact senior-value framing, bounded time update, and exact asset/risk epoch bumps. Result: 4,651 checks, 1/1 cover, 244.23s.
- `proof_v16_public_restart_empty_asset_zero_preserves_budgets_and_senior_value` calls the public zero-copy restart path with all four F histories present and both Recovery/Retired sources. It proves fresh market generation, F canonicalization, source/backing reset, budget preservation, value-stock preservation, and exact counter updates. Result: 4,282 checks, 3/3 covers, 50.77s.

This replaces the monolithic retirement proof that timed out at 300 seconds with stronger layered obligations, each under 300 seconds.

Mutation sensitivity:

- Reintroducing a current-F retirement guard falsifies arbitrary-width noninterference with both covers reached.
- Removing the long-OI blocker falsifies blocker completeness with all three covers reached.
- Mutations were restored before clean verification.

## Validation

- Production regression `v16_funded_then_flat_asset_remains_retireable`: fails with `LockActive` on the parent and passes after the fix.
- `cargo test --all-targets`: 130 passed, 0 failed.
- All four Kani harnesses above pass under 300 seconds.
- `git diff --check`: passed.
- Frozen `spec.md` unchanged.

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -12082,6 +12082,38 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
         canonical_slot
     }
 
+    #[inline(always)]
+    fn asset_has_empty_lifecycle_blocker(asset: AssetStateV16) -> bool {
+        asset.mode_long != SideModeV16::Normal
+            || asset.mode_short != SideModeV16::Normal
+            || !((asset.a_long == ADL_ONE && asset.a_short == ADL_ONE)
+                || (asset.a_long == 0 && asset.a_short == 0))
+            || asset.k_long != 0
+            || asset.k_short != 0
+            || asset.k_epoch_start_long != 0
+            || asset.k_epoch_start_short != 0
+            || asset.b_long_num != 0
+            || asset.b_short_num != 0
+            || asset.b_epoch_start_long_num != 0
+            || asset.b_epoch_start_short_num != 0
+            || asset.oi_eff_long_q != 0
+            || asset.oi_eff_short_q != 0
+            || asset.stored_pos_count_long != 0
+            || asset.stored_pos_count_short != 0
+            || asset.stale_account_count_long != 0
+            || asset.stale_account_count_short != 0
+            || asset.pending_obligation_count_long != 0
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #125, opened 2026-07-15
Keep discharged social-loss assets reusable

aeyakovenko:
## Finding

A normal public lifecycle can leave an otherwise empty asset slot permanently non-retireable:

1. Open two matched pairs on target asset 1 at a fixed price.
2. Give one target short a later asset-0 short in the same portfolio.
3. Move only asset 0, refresh the debtor, and publicly shut down asset 1.
4. The debtor's signed `ForfeitRecoveryLeg` books its cross-asset residual into asset 1's long-side B index.
5. Every target claimant exits through `ForfeitRecoveryLeg`.

At the end, target asset 1 has zero OI, position/stale/obligation counts, weights, remainders, dust, explicit loss, K, F, source backing, and reservations. Only the already-charged B history remains. The parent engine rejects public retirement with `LockActive`, permanently consuming finite market-slot capacity.

The wrapper reproduction uses only public instructions and ordinary accounts; no direct state injection creates the failure.

## Fix

Treat current and epoch-start B indices as inert history only after every claimant-bearing and value-bearing field satisfies the existing empty-state requirements. Retirement preserves the history. Restart changes `market_id` and canonicalizes the slot, so old legs cannot bind to the new market generation.

All real blockers remain enforced: side mode, ADL epoch shape, K/F, OI, position/stale/obligation counts, loss weights, social-loss remainder/dust, explicit loss, barriers, source credit, backing, reservations, and spent insurance.

## Proof-driven evidence

The source extracts the exact O(1) asset-local empty-lifecycle predicate without changing behavior, making the semantic boundary directly provable.

- `proof_v16_empty_lifecycle_classifier_is_b_history_independent` quantifies all four B lanes as arbitrary `u128` values (512 symbolic bits) across both valid ADL epoch shapes. It proves classification is identical to zero-B state. Result: 38 checks, 2/2 full-width covers, 0.109s.
- `proof_v16_empty_lifecycle_classifier_rejects_every_asset_claimant_and_loss_field` exhausts all 14 asset-local blocker classes on both sides while maximum/nonzero B history coexists. Result: 13 checks, 3/3 covers, 0.066s.
- `proof_v16_retire_empty_asset_is_value_neutral_and_epoch_scoped` calls the real zero-copy retirement path with all four B histories present. It proves success, history preservation, exact senior-value framing, bounded time update, and exact asset/risk epoch bumps. Result: 4,651 checks, 1/1 cover, 246.46s.
- `proof_v16_public_restart_empty_asset_zero_preserves_budgets_and_senior_value` calls the public zero-copy restart path with all four B histories present and both Recovery/Retired sources. It proves fresh market generation, B canonicalization, source/backing reset, budget preservation, value-stock preservation, and exact counter updates. Result: 4,282 checks, 3/3 covers, 50.39s.

This replaces the prior 362.7-second monolithic proof with stronger layered obligations, each under 300 seconds.

Mutation sensitivity:

- Reintroducing a current-B retirement guard falsifies arbitrary-width noninterference with both covers reached.
- Removing the long-OI blocker falsifies blocker completeness with all three covers reached.
- Mutations were restored before clean verification.

## Validation

- Public lifecycle regression `v16_retire_and_restart_empty_asset_after_discharged_social_loss_history`: parent fails with `LockActive`; fixed branch passes.
- `cargo test --all-targets`: 130 passed, 0 failed.
- All four Kani harnesses above pass under 300 seconds.
- `git diff --check`: passed.
- Frozen `spec.md` unchanged.

Wrapper TDD branch: `codex/probe-discharged-b-history-20260715`.

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -12121,6 +12121,10 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
             long_spent != 0 || short_spent != 0
         };
 
+        // B indices are already-charged social-loss history, not outstanding liabilities. Once all
+        // position, stale, obligation, weight, remainder/dust, explicit-loss, barrier, source,
+        // backing, and reservation state below is empty, no account can still settle either the
+        // current or epoch-start B anchor. A restart changes market_id and canonicalizes the slot.
         if slot.pending_domain_loss_barrier_long.get() != 0
             || slot.pending_domain_loss_barrier_short.get() != 0
             || asset.mode_long != SideModeV16::Normal
@@ -12135,10 +12139,6 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
             || asset.k_epoch_start_short != 0
             || asset.f_epoch_start_long_num != 0
             || asset.f_epoch_start_short_num != 0
-            || asset.b_long_num != 0
-            || asset.b_short_num != 0
-            || asset.b_epoch_start_long_num != 0
-            || asset.b_epoch_start_short_num != 0
             || asset.oi_eff_long_q != 0
             || asset.oi_eff_short_q != 0
             || asset.stored_pos_count_long != 0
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #126, opened 2026-07-15
[HARDENING] Keep recovery-asset closes local

aeyakovenko:
## What changed

- classify an outstanding close on the exact frozen Recovery asset as asset-local work
- add an engine-selected `AdvanceRecoveryClose` auto-crank step that books one configured residual chunk
- preserve global Recovery for expired closes on assets that can still drift
- allow the owner forfeit route to resume an existing close ledger
- share one O(1) close-expiry isolation kernel between production classification and expiry dispatch

## Root cause and impact

An untrusted permissionless asset authority could force its own asset into Recovery after starting a partial bankrupt close. Once the close expired, the public auto-crank classified it as `DeclareRecovery` and committed global market Recovery, blocking unrelated base-market users.

The asset mark is already frozen in `AssetLifecycle::Recovery`, so slot expiry does not require a market-wide terminal transition. The residual can continue locally under the existing close ledger and domain barrier.

## Proof-driven validation

`proof_v16_expired_frozen_asset_close_routing_is_local_and_fail_closed` symbolically covers every combination of:

- active/inactive close and all 256 bounded residual values
- expired/unexpired deadline
- frozen Recovery/drifting asset
- overlapping stale, B-stale, and liquidatable continuations

It proves every outstanding frozen-asset close selects `AdvanceRecoveryClose`, every expired drifting close selects `DeclareRecovery`, and no other state invents global Recovery. Result: 186 checks passed, 3/3 covers, 0.10s.

Mutation sensitivity: removing the frozen-asset exclusion from the production kernel fails the theorem on the local-continuation assertion while all covers remain reachable.

Additional validation:

- `proof_v16_expired_close_progress_declares_recovery_without_value_mutation`: 4,498 checks, 1/1 cover, 89.49s
- `contract_check_select_auto_crank_plan`: 463 checks, 4.14s
- `contract_check_kernel_bresidual_step`: 147 checks, 0.22s
- `git diff --check`

The frozen `spec.md` is unchanged by the final PR. A separate wrapper PR carries the public LiteSVM red/green exploit regression.


aeyakovenko:

PR #127, opened 2026-07-15
Skip Recovery legs in auto-crank selection

aeyakovenko:
## What changed

- Select the first `Active` or `DrainOnly` leg for auto-crank refresh/liquidation instead of the first active portfolio leg regardless of asset lifecycle.
- Require a dispatchable leg before classifying an account liquidatable, so Recovery-only accounts cannot advertise a fake asset-0 continuation.
- Expose `build_auto_crank_plan` so wrappers authenticate the observation for the exact engine-selected continuation without duplicating lifecycle logic.
- Route lifecycle dispatchability and liquidation eligibility through the same small production kernels used by the broad Kani theorem.

## Root cause and impact

`build_actionable_summary` correctly marked a mixed-lifecycle portfolio stale, but the auto-crank selected its first portfolio leg even when that asset was in `Recovery`. Refresh then called `accrue_asset_to_not_atomic`, which rejects Recovery assets with `LockActive`, so the sole permissionless crank could not make the available progress on a later live leg.

A Recovery-only active bitmap could also make the old classifier advertise liquidation and default to asset 0 even though no leg was dispatchable.

This is a public-state partial DoS and a contradiction of the engine's dispatch-reachability claim. It is not classified as total DoS/LoF because the Recovery asset retains its frozen-mark exit and delayed force-close routes.

## Proof-driven validation

The broad theorem `proof_v16_recovery_legs_cannot_starve_dispatchable_auto_crank_work` symbolically covers every ordering and multiplicity of disjoint Recovery, Active, and DrainOnly masks across all 16 portfolio slots. It proves:

- Recovery legs are never ordinary refresh/liquidation targets.
- The first dispatchable Active/DrainOnly leg is selected regardless of Recovery-leg placement.
- Refresh and liquidation plans carry exactly that engine-selected leg.
- Recovery-only portfolios produce no fake liquidation target.
- All five important state classes are reachable.

Result: 200 checks, 5/5 covers, 0.66s.

Mutation sensitivity:

- Treating `Recovery` as dispatchable fails the lifecycle-equivalence assertion.
- Ignoring the dispatchable-leg requirement for liquidation fails the Recovery-only assertion while 4/5 covers remain reachable.

Additional validation:

- `cargo test --all-targets`: 130 passed (50 + 9 + 2 + 69).
- `cargo fmt --all`
- `git diff --check`
- Frozen `spec.md` unchanged.
- Public wrapper LiteSVM TDD: failed with `EngineLockActive` before the fix and passes after it, refreshing the later live leg under the crank CU limit.

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -1532,6 +1532,25 @@ impl V16Core {
         None
     }
 
+    /// Ordinary refresh accrual and liquidation may target only a live or
+    /// draining asset. Recovery legs remain exit obligations, but selecting one
+    /// for an ordinary action would deterministically fail.
+    fn kernel_auto_crank_lifecycle_dispatchable(lifecycle: AssetLifecycleV16) -> bool {
+        matches!(
+            lifecycle,
+            AssetLifecycleV16::Active | AssetLifecycleV16::DrainOnly
+        )
+    }
+
+    fn kernel_auto_crank_liquidatable(
+        live: bool,
+        cert_current: bool,
+        certified_liq_deficit: u128,
+        dispatchable_asset: Option<usize>,
+    ) -> bool {
+        live && cert_current && certified_liq_deficit != 0 && dispatchable_asset.is_some()
+    }
+
     /// PRODUCTION KERNEL (engine.md selection semantics): from the actionable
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #128, opened 2026-07-15
[REAL DoS] Prevent zero-delta oracle crank pinning

aeyakovenko:
## Impact

A permissionless caller can indefinitely pin an exposed market away from its committed oracle target when the per-slot percentage cap rounds below one price atom.

For example, at price 100 with a 24 bps/slot cap, the public caller cranks every slot. Each call previously succeeded and advanced `slot_last`, but integer truncation left `effective_price` at 100. Repeating the call erased elapsed time forever, preventing the honest five-slot interval needed to move one atom toward target 200. This can freeze price-dependent PnL, liquidation, and funding progress on a live initialized market.

## Fix

When an asset is exposed, has a lagged raw oracle target, and a positive accrual segment proposes no effective-price change, return `NonProgress` before any mutation. SVM rollback then preserves `slot_last`, so elapsed time accumulates until the target-directed clamp admits a representable move.

The wrapper already applies its authenticated bounded market-accrual stage before invoking auto-crank. `AutoCrankWorkV16::observations_preaccrued` now records that trusted same-transaction fact, allowing auto-crank to dispatch the selected account refresh/liquidation without applying a second market segment. This preserves one-call account progress while ensuring each observation consumes at most one accrual segment.

The frozen `spec.md` remains unchanged. The engine accepts the pre-accrual flag only as an explicit wrapper contract: the wrapper must set it after successful authenticated pre-accrual in the same transaction and propagate every later engine error to the instruction boundary so SVM rollback covers all not-atomic mutations.

## Coverage

- Engine regression builds real OI through deposit and trade, then proves a zero-delta segment rejects byte-stably and a representable move succeeds.
- Auto-crank regression proves a pre-accrued observation refreshes the selected account while leaving the engine asset byte-identical, including when target lag remains.
- Kani proves the zero-delta lagged-target path returns `NonProgress` while preserving all relevant market and asset state.
- Public-interface LiteSVM TDD coverage and wrapper integration are in [percolator-prog#246](https://github.com/aeyakovenko/percolator-prog/pull/246), pinned to `a8b8877e784e34fbef7bfd16ce7228130219904b`.

The previous liveness model treated `slot_last` advancement as progress. It did not include target-distance progress or the integer-quantization branch, so the successful zero-delta transition satisfied the modeled rank while making the real oracle route permanently attacker-resettable.

## Verification

- `cargo test --all-targets`
- `cargo test --test v16_spec_tests v16_lagged_oracle_target_rejects_zero_delta_slot_consumption`
- `cargo kani --tests --features fuzz --harness proof_v16_lagged_target_zero_delta_cannot_consume_accrual_segment`
- `cargo fmt --all`
- `git diff --check`

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1312,10 +1312,6 @@ If full B settlement is too large, partial settlement is allowed. While `B_remai
 
 `accrue_asset_to(asset, now_slot, effective_price, funding_rate)` requires Active/DrainOnly live mode, authenticated time, valid price, and bounded funding rate. Domain locks do not block K/F/price/time accrual. Accrual MUST NOT mutate B, A, OI, weights, staged residuals, staged insurance, ADL, pending barriers, pending obligations, or exposure-clear state for a locked domain unless held by the close/recovery path.
 
-For an exposed asset with `raw_oracle_target_price != effective_price`, a positive accrual segment MUST NOT successfully consume elapsed time when the submitted effective price is unchanged. If cap quantization admits no representable price atom, accrual MUST return `NonProgress` before mutation so transaction rollback preserves the elapsed segment for a later target-directed move. Repeated permissionless calls MUST NOT reset `slot_last` while leaving the lagged effective price unchanged.
-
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #129, opened 2026-07-16
Make K/F settlement invariant to crank cadence

aeyakovenko:
## Impact

A permissionless caller could refresh a victim after every sub-atom K/F move. Each refresh rounded against the account and advanced its snapshots, so ten public cranks charged a 0.001-lot short 10 atoms while one crank over the identical mark interval charged 1 atom. Funding accrual also divided before applying index precision, allowing frequent cranks to erase otherwise nonzero funding.

The public-interface LiteSVM reproductions are in [percolator-prog PR #247](https://github.com/aeyakovenko/percolator-prog/pull/247). They use fresh valid markets, public trades, and permissionless cranks without mutating account bytes.

## Fix

- carry persistent Euclidean K and F remainders per leg;
- preserve carries across same-side resize and other leg transforms;
- apply funding index precision before settlement rounding;
- bump the on-chain layout discriminator to 18;
- leave the frozen `spec.md` unchanged.

## Broad proof-driven evidence

`proof_v16_production_kf_settlement_is_invariant_to_crank_partition` invokes the production leg-settlement transition twice versus once. It symbolically covers long/short orientation, independently active K and F lanes, signed two-segment moves including direction reversals, and nonzero persistent carries. It proves identical final K/F snapshots and remainders, identical aggregate funding and net settlement, and framing of every unrelated leg field.

`proof_v16_k_pair_carry_is_reference_exact_and_partition_invariant` independently proves the wide signed carry arithmetic against a mathematical floor/division reference over symbolic basis, denominator, initial carry, and signed partitions. It proves fragmented settlement equals delayed settlement and the canonical remainder remains bounded.

The production theorem and arithmetic theorem are complementary: the first binds the invariant to the deployed zero-copy transition; the second covers the arithmetic fallback without assuming the production fast-path shape. Removing the persistent carry or restoring per-call rounding falsifies partition invariance. The engine and public LiteSVM red/green regressions demonstrate the original cadence-dependent loss.

## Validation

- `cargo test --all-targets`
- `cargo test --features fuzz`
- `cargo kani --tests --features fuzz --harness proof_v16_production_kf_settlement_is_invariant_to_crank_partition`
- `cargo kani --tests --features fuzz --harness proof_v16_k_pair_carry_is_reference_exact_and_partition_invariant`
- Kani contracts for attach, same-side resize, and B-snapshot advance


aeyakovenko:

PR #133, opened 2026-07-16
Preserve accrual slots without observations

aeyakovenko:
## Bug

A permissionless auto-crank with no selected-asset observation could use the committed price for the account action and also advance that asset's positive-time accrual slot. A stale crank landing before the legitimate Hybrid observation could therefore consume the slot and suppress that observation; repeated permissionless landing-order races could pin Hybrid oracle progress.

The public wrapper LiteSVM reproduction uses only initialized markets/accounts and public instructions; it does not mutate protocol state out of band.

## Fix

- Preserve account-action progress using the committed price when the selected observation is absent.
- Authorize post-action market accrual iff the selected observation was supplied and the wrapper did not already pre-accrue it.
- Keep supplied-observation bounded accrual behavior unchanged.
- Leave the frozen `spec.md` unchanged.

## Proof-driven evidence

The native public-entrypoint TDD witnesses exercise stale-account refresh and liquidation. Before the fix, each omitted-observation call advanced the selected asset slot; after the fix, the account action progresses while the complete selected asset state remains unchanged.

The added Kani coverage is over production policy and the real production crank/accrual transition, not a modeled implementation clone:

- `proof_v16_auto_crank_accrual_requires_selected_observation_and_no_preaccrual` exhausts all four observation/pre-accrual states and proves the authorization biconditional. Result: 1 check, 4/4 covers, 0.026s.
- `proof_v16_observation_free_crank_preserves_asset_while_account_progresses` runs production refresh/certification/crank validation over both one-slot and multi-slot backlogs. It proves account progress, exact byte-for-byte asset-slot preservation, and unchanged vault/capital/insurance stocks. Result: 13,526 checks, 2/2 covers, 208.76s.
- `proof_v16_observed_crank_accrues_asset_while_preserving_value` proves the dual: a supplied observation consumes exactly one configured bounded segment, updates the effective price, reports remaining backlog correctly, and preserves value stocks. Result: 13,526 checks, 2/2 covers, 288.18s.

Mutation sensitivity:

- Restoring the original policy (missing observation authorizes post-action accrual) falsifies the broad omission theorem after 13,526 checks with both covers reached; the selected asset slot changes. Verification: failed as expected in 264.08s.
- Restoring the original policy and, separately, forcing all post-action accrual to skip each falsifies the four-state authorization theorem in about 0.03s.
- All mutations were restored before the clean runs and commit.

## Validation

- `cargo test --all-targets`: 132 passed, 0 failed.
- `git diff --check`: passed.
- All added Kani harnesses terminate within 300 seconds.

Stacked on the pre-accrual API in the parent branch.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1316,8 +1316,6 @@ For an exposed asset with `raw_oracle_target_price != effective_price`, a positi
 
 If the wrapper authenticates and applies the bounded market-accrual stage before calling `permissionless_auto_crank_not_atomic`, it MUST set `observations_preaccrued`. In that mode the engine dispatches the selected current-state account action without applying a second accrual segment. The wrapper MUST set the flag only after successful pre-accrual in the same transaction and MUST propagate every engine error so SVM rollback covers all not-atomic mutations.
 
-When the selected account action is realizable from committed state but its matching observation is absent, auto-crank MAY dispatch that account action but MUST NOT advance the selected asset's accrual segment. A synthesized committed price is sufficient for account refresh or liquidation bookkeeping; it is not an authenticated observation and cannot consume elapsed market time.
-
 Before any accrual/effective-price/K/F write is usable by a favorable action:
 1. affected source-domain claim-bound buckets MUST be recomputed conservatively;
 2. affected backing freshness buckets MUST be applied;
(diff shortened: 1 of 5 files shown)


aeyakovenko:

PR #134, opened 2026-07-17
[REAL LoF] Prevent ADL basis reissue across split trades

### LINKED ISSUE #131: After liquidation it is still possible to use opposing portfolio's whole `leg.basis_pos_q` by splitting it into multiple trades
`TradeNoCpi` doesn't scale the available `leg.basis_pos_q` amount by A-factor. This allows for trades made after liquidation of opposite side to be performed for whole leg amount by splitting the trade into a few smaller ones, small enough so they can pass under side OI limit.

Example scenario:

1. Perform deposits for 3 portfolios: long, short and a hacker.
2. Long and short open a trade for 100k
3. Long account accrues some maintenance fee and gets liquidated by `close_q = 25k`. `asset.a_short` becomes 75%
4. Short account attempts to trade their existing position to the hacker for 100k, but fails
5. Instead, short accounts does two trades by 50k and succeeds.

The resulting hacker account now has `a_basis` 75%, the same as asset's `a_short` 75%, so it is not supposed to get reduced in any subsequent operations. This effectively leaves asset with both OIs as 75k, while one long leg is 75k vs its short leg 100k.

<details><summary>Test for v16_spec_tests.rs</summary>
<p>

```rust
#[test]
fn v16_regression_trade_preflight_split_trades_total_oi() {
    let (mut header, mut markets) = market_fixture(1, 1);
    header.config.initial_margin_bps = V16PodU64::new(500);
    header.config.maintenance_margin_bps = V16PodU64::new(250);
    header.config.max_price_move_bps_per_slot = V16PodU64::new(100);
    header.config.max_trading_fee_bps = V16PodU64::new(10);

    let mut short_header = account_fixture(1, 49);
    let mut long_header = account_fixture(1, 50);
    let mut close_counterparty_header = account_fixture(1, 51);

    let raw_leg_basis_q =  5_451 * POS_SCALE + 9_451 * POS_SCALE;
    let liquidation_close_q = raw_leg_basis_q / 4;
    let adl_effective_oi_q = raw_leg_basis_q - liquidation_close_q;

    {
        let mut market = MarketGroupV16ViewMut::new(&mut header, &mut markets);
        let mut short = PortfolioV16ViewMut::new(&mut short_header);
        let mut long = PortfolioV16ViewMut::new(&mut long_header);
        let mut close_counterparty = PortfolioV16ViewMut::new(&mut close_counterparty_header);
        market.deposit_not_atomic(&mut short, 1_000).unwrap();
        market.deposit_not_atomic(&mut long, 1_000).unwrap();
        market
            .deposit_not_atomic(&mut close_counterparty, 1_000)
            .unwrap();
        market
            .execute_trade_with_fee_loss_stale_scoped_not_atomic(
                &mut long,
                &mut short,
                TradeRequestV16 {
                    asset_index: 0,
                    size_q: signed_q(raw_leg_basis_q),
                    exec_price: 1,
                    fee_bps: 0,
                },
            )
            .unwrap();
    }

    {
        let mut market = MarketGroupV16ViewMut::new(&mut header, &mut markets);
        let mut long = PortfolioV16ViewMut::new(&mut long_header);
        market
            .sync_account_fee_to_slot_not_atomic(&mut long, 100, 10_000)
            .unwrap();
        assert_eq!(long.header.capital.get(), 0);
        market
            .liquidate_account_not_atomic(
                &mut long,
                LiquidationRequestV16 {
                    asset_index: 0,
                    close_q: liquidation_close_q,
                    fee_bps: 0,
                },
            )
            .unwrap();
    }

    let asset = markets[0].engine.asset.try_to_runtime().unwrap();
    let short_leg = short_header.legs[0].try_to_runtime().unwrap();
    assert_eq!(short_leg.basis_pos_q, -(raw_leg_basis_q as i128));
    assert_eq!(asset.oi_eff_short_q, adl_effective_oi_q);
    assert_eq!(asset.a_short, 750_000_000_000_000);

    let mut market = MarketGroupV16ViewMut::new(&mut header, &mut markets);
    let mut short = PortfolioV16ViewMut::new(&mut short_header);
    let mut close_counterparty = PortfolioV16ViewMut::new(&mut close_counterparty_header);

    let err = market
        .execute_trade_with_fee_loss_stale_scoped_not_atomic(
            &mut short,
            &mut close_counterparty,
            TradeRequestV16 {
                asset_index: 0,
                size_q: signed_q(raw_leg_basis_q / 2),
                exec_price: 1,
                fee_bps: 0,
            },
        )
        .map(|_| ());
    assert_eq!(
        err,
        Ok(()),
        "trade succeds due to short having that amount in their position"
    );

    //////////////////////// second trade ////////////////////////
    {
        let err = market
            .execute_trade_with_fee_loss_stale_scoped_not_atomic(
                &mut short,
                &mut close_counterparty,
                TradeRequestV16 {
                    asset_index: 0,
                    size_q: signed_q(raw_leg_basis_q / 2),
                    exec_price: 1,
                    fee_bps: 0,
                },
            )
            .map(|_| ());
        assert_eq!(
            err,
            Ok(()), ///
            "trade preflight rejects closing the raw 14_902_000_000 short because \
            the side only has 11_176_500_000 ADL-effective short OI remaining"
        );
    }
}

```

</p>
</details>

aeyakovenko:
## Impact

After quantity ADL, a winning leg retained its pre-ADL raw basis while the side's effective OI and `A_side` were reduced. The public trade path could move that raw basis into a fresh portfolio in multiple fills. Each fresh leg snapped the reduced `A_side` as its new `a_basis`, reissuing previously haircutted exposure. A later mark move could therefore create more winner PnL than matched counterparty loss and consume honest backing.

The same accounting gap affected untouched post-ADL positions: K/F accrual used `ADL_ONE`, so unequal side factors produced unequal aggregate claims.

## Fix

- Scale each side's K and F index increments by that side's current A factor.
- Reject attach, flip, and same-side enlargement while either side has a non-unit A factor.
- Preserve same-side reduction and full-close liveness.
- Put the route gate in the common leg-mutation path so attach cannot rely only on an outer preflight.
- Keep the frozen `spec.md` unchanged.

## Production regressions

- `v16_quantity_adl_blocks_fresh_basis_reissue_across_split_trades` reproduces issue #131 after a quantity-ADL factor gap. The parent accepts the fresh split-transfer leg; this branch returns `LockActive` before mutation.
- `v16_quantity_adl_price_and_funding_accrual_remain_zero_sum` covers price and funding after quantity ADL, then executes a matched reduction and checks combined account value conservation.

The paired wrapper PR adds a public LiteSVM red/green exploit without direct state mutation: aeyakovenko/percolator-prog#250.

## Generalized proof obligations

The Kani suite now proves a wider ADL conservation theorem over production kernels and zero-copy views:

- Persisted admission is an exact biconditional over every lifecycle, both side modes, the full valid `A_long/A_short` range, and risk-increasing versus reducing changes.
- Route admission covers attach, flip, enlarge, reduce, and clear. Every fresh-risk route requires unit A factors, while reduce and clear remain live.
- Accrual matches the exact side-scaled K/F formulas for every valid persisted A factor.
- Cross-side K/F accrual is zero-sum for every one of the ten protocol factor quanta, including both min/max orientations.
- Production leg settlement is invariant to splitting each matched side across two accounts and remains market-level zero-sum under unequal factors.

All covers are satisfied. The production settlement theorem invokes the real leg-settlement path four times rather than proving a standalone arithmetic model.

## PDD evidence

- Mutating accrual back to `ADL_ONE` failed all four exact K/F equations in 2.04s with 3/3 covers satisfied.
- Replacing route classification with a magnitude-only gate failed the route biconditional in 14.45s; 5/6 covers remained, with the missing cover identifying the smaller side-flip bypass.

## Validation

- persisted gate: 15.15s, 7/7 covers
- route-complete gate: 14.59s, 6/6 covers
- all-valid-factor accrual: 15.23s, 3/3 covers
- factor-quantum zero-sum: 8.05s, 3/3 covers
- account-partition settlement: 213.13s, 3/3 covers
- `git diff --check`
- `spec.md` has no diff from `master`

Fixes #131.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1342,6 +1342,8 @@ If `W == 0`, residual clears only by reserved insurance or explicit protocol-own
 
 Quantity ADL applies exactly once after residual durability and is atomic with closing exposure clear/finalization or protected by a non-preemptible finalization barrier.
 
+After quantity ADL, each side's subsequent K and F index increments MUST be scaled by that side's current `A_side`; aggregate price and funding settlement across matched effective OI remains zero-sum. While either side has `A_side != ADL_ONE`, trades may reduce matched risk but MUST NOT attach, flip, or enlarge a leg. This prevents a fresh `a_basis` from reissuing quantity already removed by ADL while preserving bilateral exit progress.
+
 `begin_full_drain_reset(asset, side)` requires zero OI, no close/pending barrier, no pending obligations, no liens, and side not already `ResetPending`.
 
 -------------------------------------------------------------------------------
--- a/spec.md
+++ b/spec.md
@@ -1342,8 +1342,6 @@ If `W == 0`, residual clears only by reserved insurance or explicit protocol-own
 
 Quantity ADL applies exactly once after residual durability and is atomic with closing exposure clear/finalization or protected by a non-preemptible finalization barrier.
 
-After quantity ADL, each side's subsequent K and F index increments MUST be scaled by that side's current `A_side`; aggregate price and funding settlement across matched effective OI remains zero-sum. While either side has `A_side != ADL_ONE`, trades may reduce matched risk but MUST NOT attach, flip, or enlarge a leg. This prevents a fresh `a_basis` from reissuing quantity already removed by ADL while preserving bilateral exit progress.
-
 `begin_full_drain_reset(asset, side)` requires zero OI, no close/pending barrier, no pending obligations, no liens, and side not already `ResetPending`.
 
 -------------------------------------------------------------------------------
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -764,6 +764,26 @@ impl V16Core {
     }
 
     #[inline]
+    pub(crate) fn kernel_adl_scaled_accrual_index_deltas(
+        price_delta: i128,
+        funding_index_delta: i128,
+        a_long: u128,
+        a_short: u128,
+    ) -> V16Result<(i128, i128, i128, i128)> {
+        let a_long = i128::try_from(a_long).map_err(|_| V16Error::ArithmeticOverflow)?;
+        let a_short = i128::try_from(a_short).map_err(|_| V16Error::ArithmeticOverflow)?;
+        let k_long = checked_i128_mul(price_delta, a_long)?;
+        let k_short = checked_i128_mul(price_delta, a_short)?
+            .checked_neg()
+            .ok_or(V16Error::ArithmeticOverflow)?;
+        let f_long = checked_i128_mul(funding_index_delta, a_long)?
+            .checked_neg()
+            .ok_or(V16Error::ArithmeticOverflow)?;
+        let f_short = checked_i128_mul(funding_index_delta, a_short)?;
+        Ok((k_long, k_short, f_long, f_short))
+    }
+
+    #[inline]
     fn liquidation_progress_from_scores(before: RiskScoreV16, after: RiskScoreV16) -> bool {
         after.strictly_reduces_from(before)
             || after.certified_liq_deficit < before.certified_liq_deficit
@@ -1394,6 +1414,19 @@ impl V16Core {
             PositionRouteV16::Flip
         } else {
             PositionRouteV16::Resize
+        }
+    }
+
+    #[inline]
+    pub(crate) fn kernel_position_route_requires_unit_adl(
+        route: PositionRouteV16,
+        current: i128,
+        new: i128,
+    ) -> bool {
+        match route {
+            PositionRouteV16::Attach | PositionRouteV16::Flip => true,
+            PositionRouteV16::Clear => false,
+            PositionRouteV16::Resize => new.unsigned_abs() > current.unsigned_abs(),
         }
     }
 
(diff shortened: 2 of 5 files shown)


aeyakovenko:
Independent clean-room severity confirmation for this fix:

Using the vulnerable production SBF (`percolator-prog@da64be6`, engine `143e68c`) and only public LiteSVM instructions, a paired control/exploit test measured:

- unsplit old-basis control: 25,000 opposing loss, 50,000 backing consumed and withdrawn
- split basis-reissue path: the same 25,000 opposing loss, 100,000 backing consumed and withdrawn

The attacker therefore extracts an extra 50,000 atoms of honest external backing solely by splitting the raw-leg transfer. `ConvertReleasedPnl` and the real SPL-token `Withdraw` both succeed after the profitable claimant split-transfers its exposure away.

This confirms the PR addresses a **REAL LoF**, not only hardening. Static review also matches the exploit: the common route gate blocks fresh attach/flip/enlargement whenever either A factor is non-unit, while preserving reductions/clears, and side-scaled K/F accrual restores unequal-factor conservation.

aeyakovenko:

PR #136, opened 2026-07-18
[BLOCKER DoS] Allow Recovery market to wind down to Resolved (#236)

aeyakovenko:
## Classification: [BLOCKER DoS]

A public losing trader can organically create the outstanding close ledger that this PR was previously treating as state-injection-only.

## Public reachability

The paired wrapper LiteSVM regression uses the default certified engine profile and only public instructions:

1. Two funded users open opposite positions on a secondary asset.
2. Authenticated marks advance within the per-slot cap until one side is bankrupt.
3. The market authority performs the normal secondary-asset shutdown transition.
4. The losing owner calls `ForfeitRecoveryLeg` once. Bankruptcy booking commits one default bounded chunk and leaves a real residual close ledger, then the owner disappears.
5. Force-close, the opposite owner's forfeit/reduce paths, and pre-expiry auto-crank are blocked.
6. After the close lifetime expires, auto-crank commits global `Recovery`.

The earlier premise was incomplete: `liquidate_account_not_atomic` is not the only public close-ledger producer. `forfeit_recovery_leg_not_atomic` can begin and partially advance a bankruptcy close ledger without the liquidation preflight.

On the old engine, `resolve_market_not_atomic` rejects `Recovery`. Withdrawal requires `Live`, while `CloseResolved` and terminal cleanup require `Resolved`, so the attacker can strand every independent user.

## Fix

Allow any non-`Resolved` market (`Live` or `Recovery`) to transition to `Resolved`; reject only repeated resolution. The resolved payout path remains self-gating:

- positive payouts wait for all loss barriers,
- close progression remains bounded,
- custody movement remains checked by `TokenValueFlowProofV16`.

## Evidence

Paired wrapper PR: aeyakovenko/percolator-prog#252.

- exact wrapper parent `36fa587e1424a8c7fdde2c701c10938188a51876`
- old engine `143e68c4917ed0400a27b952f036a5677047cd84`
- RED: mature permissionless resolve returns `Custom(21)` after the public sequence reaches `Recovery`
- fixed engine backport `493fcacc55e236af36e0c8c6e683923ee50a2365`
- GREEN: market reaches `Resolved`; abandoned loser and independent winner complete bounded terminal steps; the canonical SPL vault pays the winner with exact accounting
- wrapper LiteSVM/CU suite: 566/566

Engine regression and Kani frame/value-neutrality proofs remain part of this PR.

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -8637,6 +8637,37 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
         Ok((V16PodU128::default(), V16PodU128::default()))
     }
 
+    fn clear_terminal_source_backing_pair(
+        market_id: u64,
+        source: SourceCreditStateV16,
+        bucket: BackingBucketV16,
+    ) -> V16Result<(SourceCreditStateV16, BackingBucketV16)> {
+        if source.is_empty_amount_shape() && bucket.is_empty_amount_shape() {
+            return Ok((source, bucket));
+        }
+        if source.positive_claim_bound_num != 0
+            || source.exact_positive_claim_num != 0
+            || source.fresh_reserved_backing_num != 0
+            || source.valid_liened_backing_num != 0
+            || source.impaired_liened_backing_num != 0
+            || source.insurance_credit_reserved_num != 0
+            || source.valid_liened_insurance_num != 0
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #138, opened 2026-07-18
Price oracle target lag through final margin

aeyakovenko:
## What changed

- defer target/effective price lag to the existing post-trade margin decision instead of rejecting every risk-increasing trade at preflight
- keep the preflight blocks for asset-loss-stale risk and pending domain barriers
- update the NB1 proof-only admission model so target lag is represented by the target-aware final margin guard
- add a runtime regression showing a sufficiently collateralized trade remains live while an account one atom below the target-aware requirement is rejected

## Why

Authenticated oracle targets need to become visible to engine risk checks as soon as they are accepted. The engine already invalidates health certificates on target changes, recomputes the complete account certificate when target and effective prices differ, adds the entire adverse target delta to IM/MM/worst-case loss, and checks final IM on both trade accounts.

The separate preflight target-lag block therefore prevented economically valid trades without adding protection. It also made immediate target publication unusable by the wrapper. This change preserves trade liveness while retaining the target-aware collateral boundary.

## Validation

- `cargo test --all-targets` (130 native/spec/fuzz tests passed)
- Kani `proof_v16_trade_preflight_risk_gate_blocks_only_unsafe_risk_increase`
- Kani contract `contract_check_kernel_trade_preflight_admits`
- Kani contract `contract_check_kernel_economically_valid_trade_admits`
- Kani `proof_v16_target_effective_lag_penalty_enters_all_health_lanes`

The public LiteSVM loss-of-funds repro and wrapper-side target publication are kept in a separate dependent wrapper PR.

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -16326,12 +16326,10 @@ fn position_delta_increases_risk(current: i128, delta_q: i128) -> V16Result<bool
 fn trade_preflight_risk_gate(
     risk_increasing: bool,
     asset_loss_stale: bool,
-    target_effective_lag: bool,
+    _target_effective_lag: bool,
     touches_pending_domain_barrier: bool,
 ) -> V16Result<()> {
-    if touches_pending_domain_barrier
-        || (risk_increasing && (asset_loss_stale || target_effective_lag))
-    {
+    if touches_pending_domain_barrier || (risk_increasing && asset_loss_stale) {
         return Err(V16Error::LockActive);
     }
     Ok(())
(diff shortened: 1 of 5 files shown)


aeyakovenko:
Independent real-SBF integration evidence supports this dependency.

I built wrapper `f7a64eb0759b3ec58396cadfdc6793ff2c896ed2` against this exact engine head `2af64e487f3dec88b18aac634071115e588d2af1` and ran a separately written public AuthMark exposure-migration attack. On the production parent, an attacker moves a known-losing short into a thin portfolio, withdraws the funded portfolio, gains 40,000,000 atoms, and underpays an independent long by 40,000,000. With this engine head plus immediate wrapper target publication, the thin migration rejects byte-atomically and all terminal payouts/backstop balances match the no-migration control.

Static path review matches the result: target publication increments `oracle_epoch`; lagged legs add the entire adverse target delta to IM, MM, and worst-case loss; post-fill recertification recomputes the full target-aware certificate; and final IM is checked for both accounts. This PR removes only the redundant blanket target-lag preflight, preserving the target-aware final-margin boundary and adding no value-moving or authority surface.

aeyakovenko:

PR #139, opened 2026-07-18
[REAL DoS] Reset exact-OI ADL trade residues

aeyakovenko:
## Summary

- Detect crossed trades that reduce one side's effective OI exactly to zero while ADL survivor basis remains stored.
- Move that side into the existing `ResetPending` bounded-cleanup route after all trade margin checks.
- Leave the affected certificate stale so the permissionless cleanup from engine #99 refreshes and removes the dead leg.

## Public red case

Paired wrapper LiteSVM TDD: https://github.com/aeyakovenko/percolator-prog/pull/266

Through ordinary public instructions, two users open opposing positions, one side is partially liquidated, and a crossed trade reduces effective OI by exactly the remaining amount. On the pre-fix dependency head, the trade succeeds but leaves `1_000_000` basis units in an active survivor leg while both effective-OI counters are zero and both sides remain `Normal`.

The survivor's direct reduction returns `CounterUnderflow`, another willing counterparty cannot trade the position out, and ten permissionless cranks each return success while making no state change. Deposited capital remains locked despite an honest cranker.

The strict-greater-than case was fixed by #109. This is the adjacent equality boundary and a different production route from engine #118 / wrapper #219, which concern unilateral bankruptcy close.

## Fix

After successful trade checks, inspect only touched assets. If a side has zero effective OI, still has stored position basis, has no pending payout obligation, and is not already resetting, start the existing full-drain reset. This preserves trade availability and delegates cleanup to the bounded permissionless route.

## Validation

- rebased public red `bc790f69`: OI `(0,0)`, modes `(Normal,Normal)`, residual `1_000_000`; owner exit `Custom(25)`; 10/10 successful cranks are identical no-ops
- fixed public head `2fe02df1`: enters `ResetPending`; one crank clears the leg at 100,151 CU; all remaining principal withdraws
- `cargo test --all-targets --features fuzz`: 161 passed
- Kani `proof_v16_trade_reductions_are_funded_only_by_preexisting_side_oi`: passed
- Kani `proof_v16_prior_reset_cleanup_cannot_starve_live_liquidation`: passed
- paired wrapper aggregate: 6 unit + 568 serialized LiteSVM/CU tests passed
- `git diff --check`

## Stack

Base: engine PR #99 at `28a7338bd8c0dff7134ea0aa06404795bbb8557b`.

Head: `619f3fec9a2f6c0cf2462414c47a65eeefdde81c`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Source-checkable counts in this checkout:
 
 | Class | Count |
 | --- | ---: |
-| Plain Kani proofs in `tests/proofs_v16.rs` | 244 |
+| Plain Kani proofs in `tests/proofs_v16.rs` | 247 |
 | Plain Kani proofs in `tests/proofs_v16_arithmetic.rs` | 11 |
 | Plain Kani proofs in `src/v16_proofs.rs` | 38 |
 | Function-contract proofs in `src/v16_proofs.rs` | 51 |
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -14092,10 +14092,6 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
             short_delta,
             trade_preflight.short_lookup,
         )?;
-        // ADL can leave stored basis larger than effective OI. If a matched trade consumes the
-        // final effective unit while basis remains, enter the existing reset epoch so public crank
-        // can clear the economically exhausted residue instead of leaving a Normal-mode zero-OI leg.
-        self.begin_zero_oi_trade_residue_resets(request.asset_index)?;
         self.transfer_trade_residual_reward_credit(
             long_account,
             short_account,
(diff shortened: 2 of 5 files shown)


aeyakovenko:
Added the missing generalized Kani continuation theorem in f3306c57. It invokes the exact mutable-view reset helper over symbolic long/short OI, stored counts, pending obligations, all side modes, epochs, K/F/B state, loss weights, remainders, and dust. It proves complete independent ResetPending routing, exact epoch/risk-epoch changes, social-loss quarantine conservation, global value neutrality, full unrelated-header/slot framing, and the no-op branch.

Verification:
- Kani: 3,675 checks passed, 6/6 covers, 85.5s
- `cargo test --quiet`: passed
- Public exact-OI trade regression: passed
- Mutation checks: omitting the short-side reset fails the theorem; disconnecting the helper from the trade path fails the public regression.

Production source is unchanged; the only accessor is cfg(Kani/fuzz).

aeyakovenko:
Superseded by consolidated engine dependency #195. Every non-merge patch unique to #139 is patch-identical to a commit retained in #195 head 495a5590c97055bd71c6f94d849ff0298f243145.

PR #140, opened 2026-07-18
[REAL LoF] Keep source-claim conversion domain-local

aeyakovenko:
## Root cause

Account-wide source conversion selected backing by source domain, but returned only an aggregate face burn. The later PnL setter retired source claims in portfolio slot order. A winner with claims in two domains could therefore consume domain 3 backing while burning domain 1 claim, then consume domain 3 backing again. This crosses asset-isolation boundaries and overdraws an independent backing provider.

## Fix

- Retire each selected source claim in the same domain whose backing is consumed.
- Pass the exact pre-burned bound numerator into the aggregate PnL setter so it cannot double-burn claim face.
- Reject live conversion while any source claim remains liened.

## TDD

- `e7466b4` is the test-only commit and fails on the unmodified engine because domain 1 disappears while domain 3 survives.
- `f0e189a` is fixed-green.
- A dependent wrapper PR supplies the public LiteSVM exploit sequence without state injection.

## Verification

`cargo test --all-targets` passes: 50 unit/math, 9 backing fuzz, 2 resolved-insolvency fuzz, and 69 V16 spec tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Source-checkable counts in this checkout:
 
 | Class | Count |
 | --- | ---: |
-| Plain Kani proofs in `tests/proofs_v16.rs` | 244 |
+| Plain Kani proofs in `tests/proofs_v16.rs` | 245 |
 | Plain Kani proofs in `tests/proofs_v16_arithmetic.rs` | 11 |
 | Plain Kani proofs in `src/v16_proofs.rs` | 38 |
 | Function-contract proofs in `src/v16_proofs.rs` | 51 |
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -14599,10 +14599,12 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
         if released == 0 {
             return Ok(0);
         }
-        if Self::account_has_source_claims(&account.as_view())?
-            && self.account_has_active_source_claim_exposure(&account.as_view())?
-        {
-            return Err(V16Error::LockActive);
+        if Self::account_has_source_claims(&account.as_view())? {
+            if self.account_has_active_source_claim_exposure(&account.as_view())?
+                || Self::valid_source_lien_effective_reserved_sum(&account.as_view())? != 0
+            {
+                return Err(V16Error::LockActive);
+            }
         }
         let converted = if Self::account_has_source_claims(&account.as_view())? {
             self.account_source_realizable_support(&account.as_view(), released)?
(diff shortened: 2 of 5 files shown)


aeyakovenko:
Public-interface LiteSVM TDD and wrapper pin: https://github.com/aeyakovenko/percolator-prog/pull/267. Exact-parent red overdraws domain 3 backing 200 atoms for a 100-atom domain-local claim; fixed-green caps it at 100 under the 300k custody CU limit.

aeyakovenko:
Independent clean-room confirmation against the exact production SBF pinned by `percolator-meta`:

- wrapper: `percolator-prog@da64be639168b496833dd4c701607a94fcb06b6c`
- engine: `percolator@143e68c4917ed0400a27b952f036a5677047cd84`
- public LiteSVM instructions only; no initialized Percolator state mutation

I used a fresh four-slot market and one winner portfolio:

1. Open 20-lot longs on assets 0 and 1 at 100.
2. Move both authenticated marks to 105 and refresh only the winner, creating exactly 100 atoms of positive PnL claim in domains 1 and 3. The original losing counterparty stays stale, so domain 1 has zero backing.
3. Deposit 200 atoms of external backing only into domain 3.
4. Flatten the winner against a fresh replacement portfolio, preserving 200 total PnL and no active exposure.
5. Convert 100 atoms, perform a bounded same-price open/close to recertify, convert the remaining 100, then withdraw 200 through the real SPL-token path.

Observed on the vulnerable pin:

- before conversion: domain 1 claim = 100, backing = 0; domain 3 claim = 100, backing = 200
- first conversion consumes 100 domain-3 backing but retires domain 1's claim; domain 3's 100 claim survives
- second conversion consumes another 100 from domain 3
- domain 3 `consumed_liened_backing_num` and `provider_receivable_num` both end at 200 atoms for its original 100-atom domain-local claim
- the trader's capital rises from 1,000 to 1,200 and a public `Withdraw` transfers all 200 atoms to their token account

This independently confirms **REAL public LoF** and the cross-domain isolation violation.

Static review of this PR matches the root cause: it burns each selected claim in the same domain whose backing is consumed and passes the pre-burned bound numerator into the aggregate PnL update. The ceiling conversion in `amount_from_bound_num` also means the atom-denominated PnL decrease is not smaller than the pre-burned bound numerator, avoiding an obvious rounding underflow.

aeyakovenko:
Added a generalized Kani theorem in 09e76bd4. It quantifies all four Kani source domains and proves: selected-domain confinement, unrelated-domain framing, exact aggregate claim-rank decrease, disjoint source-preburn/residual-burn accounting, mismatched-domain rejection, and fail-closed over-preburn handling.

Verification:
- Kani: 476 checks passed, 4/4 covers, 74.7s
- `cargo test --quiet`: passed
- Targeted public conversion regression: passed
- Mutation checks: removing the domain comparison fails the theorem; ignoring preburn fails two theorem obligations; removing the production burn call fails the public entrypoint regression.

No spec, layout, or public API change.

aeyakovenko:

PR #141, opened 2026-07-18
[REAL DoS] Reset unilateral exact-OI ADL residues

aeyakovenko:
## What changed

After a unilateral reduction, detect either side with zero effective OI but surviving stored basis and begin the existing full-drain reset. The reset is overlap-safe with the opposite-side reset already started by unilateral close.

## Root cause and impact

ADL can make stored basis exceed effective OI. An owner can then publicly call `RebalanceReduce` for exactly the remaining effective OI. The old path reset only the opposite side, leaving the owner's side `Normal` with zero OI and nonzero basis. Rebalance, a fresh-counterparty trade, and full withdrawal all reject, while repeated honest auto-cranks return without clearing the leg, stranding deposited collateral.

## TDD coverage

The regression constructs the exact ADL accounting relation, refreshes the account, performs the unilateral reduction through the production engine entrypoint, and requires both zero-OI residue sides to enter `ResetPending`.

A companion percolator-prog PR exercises the full sequence through public LiteSVM instructions on the exact pre-fix engine parent and fixed commit. This PR is intentionally stacked on #139, which supplies the shared exact-OI reset detector and the prior reset auto-crank dependency without duplicating either fix.

## Verification

- `cargo test --test v16_spec_tests v16_exact_oi_unilateral_reduce_starts_reset_for_adl_basis_residue -- --nocapture`
- `cargo test` (133 tests passed)

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -13594,7 +13594,7 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
         Ok(())
     }
 
-    fn begin_zero_oi_trade_residue_resets(&mut self, asset_index: usize) -> V16Result<()> {
+    fn begin_zero_oi_residue_resets(&mut self, asset_index: usize) -> V16Result<()> {
         let asset = self.asset_state(asset_index)?;
         let reset_long = asset.oi_eff_long_q == 0
             && asset.stored_pos_count_long != 0
@@ -13830,6 +13830,7 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
         }
         self.reduce_position(account, request.asset_index, close_q)?;
         self.certify_account_after_local_settlement_with_price_override(account, None)?;
+        self.begin_zero_oi_residue_resets(request.asset_index)?;
         self.validate_liquidation_progress_from_score(before_score, &account.as_view())?;
         self.validate_shape_audit_scan()?;
         self.validate_account_audit_scan(&account.as_view())?;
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Independent red/green review confirms this PR fixes a **REAL public DoS** on the exact Meta production pin.

I reproduced the issue from normal public SBF instructions without using this PR's test or injecting initialized state. After partial liquidation, an exact-effective-OI `RebalanceReduce` leaves OI `(0,0)`, modes `(Normal, ResetPending)`, stored counts `(1,1)`, and live basis on both portfolios. Fresh-slot cranks, both owner reductions, a willing bilateral close, full withdrawal, and survivor forfeiture cannot clear the `Normal` residue. Even cooperation from the opposite owner only removes the opposite leg.

I then built wrapper `b9c6b64e08e4904a64286c1e46bc4eae913add22` with this engine head `b3886dcf6927c8e1cb2b822f0819451971a4b3a5` and reran the same independent sequence. Both sides enter `ResetPending`; the first bounded crank for each portfolio detaches the leg; both owners then withdraw all remaining capital.

Static review: calling `begin_zero_oi_residue_resets` after unilateral liquidation/rebalance certification covers the closed-side equality residue while preserving the existing opposite reset. The helper refuses to reset a side with pending payout obligations and skips an already pending reset. I found no added authority or value-moving surface and no conflicting accounting path in this diff.

aeyakovenko:
Rebased this stacked PR onto #139 commit f3306c57, so the unilateral/liquidation routes now inherit the generalized zero-OI continuation theorem. Updated the cfg(Kani/fuzz) accessor name to match the shared helper in 7b668126.

Verification on this head:
- generalized Kani theorem: 3,675 checks passed, 6/6 covers, 82.1s
- public unilateral exact-OI regression: passed
- fmt/diff checks: passed

The theorem is deliberately shared rather than duplicated: it proves the exact production transition for both sides and all modes; this PR’s public regression binds the unilateral entrypoint to that transition.

aeyakovenko:
Superseded by consolidated engine dependency #195. Every non-merge patch unique to #141 is patch-identical to a commit retained in #195 head 495a5590c97055bd71c6f94d849ff0298f243145.

PR #144, opened 2026-07-19
[REAL LoF] Retire B losses from their source domain

aeyakovenko:
## Impact

B/social-loss settlement reduced aggregate positive PnL in sparse portfolio-slot order instead of retiring the claim in the bankrupt asset's source domain first. A multi-asset winner could therefore burn an unrelated claim, preserve the claim backed by the affected asset, and realize that preserved claim against an independent canonical provider during terminal close.

The reproduction uses normally initialized engine state and public-equivalent transitions. Asset 0's short-source claim occupies an earlier portfolio slot, asset 1's short-source claim is backed by an independent provider, and an asset-1 bankruptcy creates B loss. Before this fix, settling B burned asset 0's claim first and left the asset-1 claim extractable.

## Root cause

`settle_account_b_chunk` computed the affected asset but delegated its positive-PnL reduction to aggregate `set_account_pnl`, whose source-claim traversal intentionally follows sparse account-slot order. That order is valid for generic aggregate reductions but not for asset-scoped B loss.

## Fix

- Retire B loss from `insurance_domain_index(asset_index, opposite_side(leg.side))` first.
- Fall back to other source claims only when the affected domain cannot cover the full reduction.
- Reuse PR #111's proved partial source-lien unwind path instead of duplicating lien arithmetic.
- Validate the already-selected physical source-claim slot after a prior slot is emptied, so the sparse terminator cannot hide a later fallback claim.
- Preserve the existing aggregate sparse-slot behavior for non-B callers.

## TDD and verification

- Exact-parent RED at `5027db0e`: `v16_b_loss_burns_the_matching_source_domain_before_sparse_slot_order` fails with the unrelated asset-0 claim erased.
- Sparse fallback regression: `v16_source_claim_burn_crosses_a_just_emptied_sparse_slot` covers an earlier claim becoming the terminator before a later physical slot is consumed.
- `cargo test --all-targets --features fuzz`: 164 tests pass.
- Re-ran the inherited source-lien partition, minimality/isolation, backing-release, and fee-proration Kani proofs: all pass.
- A public LiteSVM terminal extraction reproduction passes on the fixed wrapper integration, and the full wrapper suite passes 6 unit + 566 LiteSVM/CU tests.

## Dependency

This PR is intentionally stacked on #111 because it consumes that PR's production, contract-proved partial source-lien unwind primitive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Source-checkable counts in this checkout:
 
 | Class | Count |
 | --- | ---: |
-| Plain Kani proofs in `tests/proofs_v16.rs` | 244 |
+| Plain Kani proofs in `tests/proofs_v16.rs` | 249 |
 | Plain Kani proofs in `tests/proofs_v16_arithmetic.rs` | 11 |
 | Plain Kani proofs in `src/v16_proofs.rs` | 38 |
 | Function-contract proofs in `src/v16_proofs.rs` | 51 |
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -8844,51 +8844,11 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
             }
             let d = source_snapshot.domain.get() as usize;
             self.domain_asset_side(d)?;
-            let burnable = Self::source_claim_unliened_num(&account.as_view(), d)?;
-            let burn = burnable.min(burn_num);
-            if burn != 0 {
-                self.decrement_account_source_claim_for_domain_not_atomic(account, slot, d, burn)?;
-                burn_num -= burn;
-            }
-            if burn_num != 0 {
-                let liened_burn = account.header.source_domains[slot]
-                    .source_claim_liened_num
-                    .get()
-                    .min(burn_num);
-                if liened_burn != 0 {
-                    self.burn_account_source_lien_face_not_atomic(account, slot, d, liened_burn)?;
-                    burn_num -= liened_burn;
-                }
-            }
-            if burn_num != 0 {
-                let (impaired_burn, impaired_effective_burn) =
-                    Self::burn_impaired_account_source_claim_fields(account, slot, burn_num)?;
-                if impaired_burn != 0 {
-                    if decode_market_mode(self.header.mode)? == MarketModeV16::Resolved
-                        && impaired_effective_burn != 0
-                    {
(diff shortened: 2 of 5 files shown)


aeyakovenko:
The source-domain accounting defect is reproducible, but the current retained regression does not yet prove the PR title's terminal LoF classification.

On the current Meta integration pins (`percolator c1a058f`, `percolator-prog cf69b40`), an asset-1 B settlement does incorrectly retire asset-0 claim face first. Your fix makes that accounting assertion pass. However, the public terminal extraction attempt is blocked by a distinct confirmed DoS: after bankruptcy reduces effective OI and a later bilateral close reaches zero OI, a normal-epoch residual basis leg cannot be cleared by live, recovery, or resolved paths. The clean LiteSVM reproduction is `aeyakovenko/percolator-meta@9c75c9d`; details are on engine #118 and program #219.

That means the current test proves cross-domain liability corruption, not yet an independently backed provider's realized token loss. In addition, the SBF I built from this PR stack (SHA-256 `e5edc9df8b2dbed6ec9fc67cb26d5ee54cbd51d331c5ad68befd549d122d8495`) drops the source-capacity admission fix in `c1a058f` and reopens the public terminal DoS covered by Meta PR #275.

Please stack/rebase on the source-capacity fix and the bilateral-zero-OI repair, then retain a public LiteSVM test that reaches provider-backing consumption or another independent victim net loss. Until that terminal proof exists, this is a confirmed accounting bug / latent LoF, not a demonstrated live custody LoF.

aeyakovenko:
Added commit d57a85fa with a full-width domain-first B-loss retirement theorem over every representable two-domain claim rank. It proves exact aggregate claim conservation, frames the unrelated domain while the affected source can cover the burn, and permits fallback only after the source is exhausted; all 3 non-vacuity covers are reachable.

PDD checks:
- Aggregate-first kernel mutation fails both Kani and the sparse-order production regression.
- Dropped-fallback mutation fails exact conservation and the production fallback regression.
- Caller bypass mutation fails the production sparse-order regression.

Verification: theorem passes in 1.85s; 130 default tests and 164 `fuzz`-feature tests pass; `cargo fmt --check` and `git diff --check` pass. No spec, account-layout, or public runtime API changes.

aeyakovenko:

PR #146, opened 2026-07-20
[HARDENING] Retire terminal non-reclaimable fees safely

aeyakovenko:
## What changed

- add a terminal-only engine transition that retires unbudgeted insurance after every portfolio, PnL, backing, payout, domain-budget, and source-credit claim has cleared
- require engine vault accounting to equal the claim-free insurance amount before retirement
- expose the pure retirement delta to Kani and prove its exact value flow

## Why

Wrapper PR aeyakovenko/percolator-prog#225 keeps EWMA mark-movement fees outside all authority-withdrawable insurance budgets. Those fees must remain non-reclaimable while a market is live, but the final market account still needs a bounded way to close after all users and claims have exited.

The wrapper pairs this accounting transition with an atomic burn from the canonical SPL vault. A failed or malformed burn rolls the engine transition back at the SVM instruction boundary.

## Security boundary

This method rejects unless the market is resolved and fully claim-free. It cannot retire domain-budgeted insurance, source-credit reservations, portfolio capital, positive PnL, backing earnings, or unresolved payout receipts.

## Validation

- `cargo test --all-targets --features fuzz`: 159 passed
- `cargo kani --tests --exact --harness kani_retire_terminal_unbudgeted_insurance_delta_contract`: 132 checks passed, 3/3 covers
- wrapper exact-parent public LiteSVM regression and all `567/567` wrapper tests pass on this pin

Companion wrapper fix: aeyakovenko/percolator-prog#225

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Source-checkable counts in this checkout:
 
 | Class | Count |
 | --- | ---: |
-| Plain Kani proofs in `tests/proofs_v16.rs` | 244 |
+| Plain Kani proofs in `tests/proofs_v16.rs` | 252 |
 | Plain Kani proofs in `tests/proofs_v16_arithmetic.rs` | 11 |
 | Plain Kani proofs in `src/v16_proofs.rs` | 38 |
 | Function-contract proofs in `src/v16_proofs.rs` | 51 |
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -9196,6 +9196,74 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
         self.validate_shape()
     }
 
+    // Contract layer: terminal retirement removes only insurance that has no
+    // remaining domain or source-credit claimant. The wrapper pairs this
+    // accounting transition with an SPL-token burn.
+    #[cfg_attr(all(kani, feature = "contracts"), kani::ensures(|result: &V16Result<(u128, u128, u128)>| match result {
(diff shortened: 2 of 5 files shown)


aeyakovenko:
Proof-strengthening follow-up pushed in `b39528d7`.

The original proof covered only the pure retirement delta. The added Kani obligations bind the production zero-copy route and prove:
- claim-free retirement burns exactly the full insurance balance across the protocol vault range and frames every other header/slot field;
- account capital, positive source claims, backing-provider earnings, and fresh backing principal each independently block retirement;
- domain budgets, source insurance reservations, pending close barriers, and materialized portfolios each independently block retirement;
- Live mode, stale certificates, B-stale accounts, and negative-PnL accounts each independently block retirement;
- every rejection pre-state is shape-valid before the call and the complete persisted header/slot remains unchanged.

All new harnesses complete under 300s individually: 43s–63s for ordinary classes, 61s for full-width success, 261s for the reservation matrix, and 265s for the U256 source-claim state. All covers are satisfiable.

Mutation checks:
- Removing every positive-PnL/source-claim guard makes the source-claim theorem fail (`result != LockActive`, 247s).
- Removing the Resolved-mode guard makes the readiness theorem fail (53s).

Verification:
- `cargo test --all-targets --features fuzz --quiet`: 159 passed.
- Production source is unchanged by this follow-up.

aeyakovenko:

PR #147, opened 2026-07-20
[REAL LoF / DoS] Credit post-snapshot expired backing to payouts

aeyakovenko:
## Impact

Fresh source backing is senior while live. Once it expires after the resolved payout snapshot, those atoms become junior residual but the vulnerable engine never added them to the frozen payout ledger. Ordinary close order could therefore leave user entitlement trapped in the vault and prevent terminal value reconciliation.

The public LiteSVM regression creates two independent winner/loser pairs, captures a 40/80 payout snapshot, then expires a second 40-atom backing bucket. On the vulnerable parent both losers receive 960, both winners receive only 1,020 from four 1,000-atom deposits, and 40 atoms remain undistributable.

## Fix

- Credit residual released after payout-snapshot capture into both persisted payout snapshots.
- Recompute the common payout rate with kernel_recompute_resolved_payout_rate.
- Preserve replay inertness and reject every ineligible bucket.

## TDD evidence

- RED commit a7577b0b carries the exact-parent public regression.
- GREEN commit e35f0f6d adds the production fix and conservation proofs.

## Verification

- cargo test --all-targets --features fuzz: 181 tests pass.
- Five targeted Kani harnesses pass, covering exact post-snapshot credit, public haircut improvement, full-rate restoration, ineligible-bucket rejection, and replay inertness.
- The principal conservation theorem reports 4/4 nonvacuity covers.
- Source proof roster: 266 plain proofs in tests/proofs_v16.rs.

Stacked on #149, which prepares lapsed source state before resolved settlement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Source-checkable counts in this checkout:
 
 | Class | Count |
 | --- | ---: |
-| Plain Kani proofs in `tests/proofs_v16.rs` | 254 |
+| Plain Kani proofs in `tests/proofs_v16.rs` | 266 |
 | Plain Kani proofs in `tests/proofs_v16_arithmetic.rs` | 11 |
 | Plain Kani proofs in `src/v16_proofs.rs` | 38 |
 | Function-contract proofs in `src/v16_proofs.rs` | 51 |
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -1499,6 +1499,48 @@ impl V16Core {
         (payout, new_vault)
     }
 
+    /// Recompute the resolved payout rate from the ledger's current residual
+    /// and outstanding claim bound. This deliberately contains no wide
+    /// division: receipts apply the resulting fraction when they are paid.
+    pub(crate) fn kernel_recompute_resolved_payout_rate(
+        mut ledger: ResolvedPayoutLedgerV16,
+    ) -> V16Result<ResolvedPayoutLedgerV16> {
+        let total_bound_num = ledger
+            .terminal_claim_exact_receipts_num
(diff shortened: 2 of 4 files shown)


aeyakovenko:
Proof-strengthening follow-up pushed in `a7f0a404`.

Added/generalized obligations:
- Full-width payout conservation across symbolic released backing, junior residual, mixed receipted/unreceipted claims, snapshot metadata, and both haircut branches.
- Two production zero-copy route bindings: partial haircut improvement and restoration to full payout rate.
- Exhaustive expiry eligibility over all bucket statuses plus not-yet-expired `Fresh` state.
- Production replay frame proving a second expiry is `Stale` and cannot mutate backing, source aggregates, risk epoch, payout snapshots, vault, capital, or insurance.
- Separate pre-snapshot under-backed and fully-backed liveness routes proving expiry removes realizable source support without initializing terminal accounting.

Non-vacuity/mutation checks:
- All four symbolic payout covers are satisfiable.
- Disabling the post-snapshot credit call fails the production route theorem.
- Weakening the expiry eligibility gate fails both the status theorem and the replay frame.

Verification:
- All added Kani harnesses complete within 71s individually.
- `cargo test --all-targets --features fuzz --quiet`: 166 passed.
- Production source is unchanged by this follow-up.

aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #147 (e35f0f6d81cae87484eaf64743ce3a73520ce450) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #148, opened 2026-07-20
[REAL LoF] Preserve bankruptcy loss domains across leg exits

aeyakovenko:
## Finding

A public cross-margin sequence could realize a loss on one asset, flatten that loss-bearing leg, and leave an unrelated leg active. The aggregate negative PnL retained no source-domain provenance, so later liquidation inferred the bankruptcy domain from the unrelated surviving leg and spent that domain's insurance.

The exact-parent wrapper LiteSVM regression demonstrates an independent asset-0 insurance provider losing `100100` atoms while the attacker receives `115300` atoms and nets `95100` atoms from an asset-1 loss. It uses only normal market initialization, trades, authenticated mark updates, maintenance, liquidation, and close instructions; it does not inject account state.

## Fix

- Persist negative-PnL atoms by insurance domain in each portfolio.
- Require the per-domain total to equal aggregate negative PnL.
- Attribute every loss-producing settlement path and consume offsets in lockstep.
- Route live and resolved bankruptcy settlement through the recorded domain, including multi-domain close progression.
- In Resolved mode, record each remaining terminal debt domain as explicit loss after the payout haircut instead of adding a second social-loss charge to winners.
- Bump the portfolio layout discriminator from 17 to 18. There are no deployed markets requiring migration.

This preserves the rule that one asset's insolvency cannot consume an independent asset's insurance after the originating leg exits.

## Verification

- `cargo test --all-targets --features fuzz`: 160 tests passed.
- Native regressions cover live liquidation domain selection and resolved multi-domain rotation.
- Eleven focused Kani harnesses passed on head `ab111c7e` with zero failures and all cover goals reached:
  - unattributed negative PnL is rejected
  - resolved bankruptcy uses recorded provenance
  - domain insurance spending remains budget-bounded
  - principal settles before bankruptcy residual
  - negative PnL settles before fee extraction
  - an unfunded domain cannot consume shared insurance
  - loss accrual creates exact domain provenance inductively
  - principal reduction conserves provenance across two domains
  - a domain cure frames unrelated debt
  - sparse principal preflight agrees with the production kernel
  - sparse targeted cure binding frames the unrelated domain

The complete 246-harness Kani roster was not rerun in this PR; three obsolete or redundant harnesses were removed after the new persisted invariant made their old fixtures unreachable or duplicated an existing ordering theorem. The resolved rotation regression uses live nonzero loss weights on both domains and asserts that neither domain's B index is incremented.

## Scope

The public regression can also encounter the already-reported SVM-rollback liveness issue when `RecoveryRequired` attempts to persist recovery state and the instruction rollback erases it. That is the separate wrapper #190 / engine #114 issue and is not duplicated here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Source-checkable counts in this checkout:
 
 | Class | Count |
 | --- | ---: |
-| Plain Kani proofs in `tests/proofs_v16.rs` | 241 |
+| Plain Kani proofs in `tests/proofs_v16.rs` | 246 |
 | Plain Kani proofs in `tests/proofs_v16_arithmetic.rs` | 11 |
 | Plain Kani proofs in `src/v16_proofs.rs` | 38 |
 | Function-contract proofs in `src/v16_proofs.rs` | 51 |
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Source-checkable counts in this checkout:
 
 | Class | Count |
 | --- | ---: |
-| Plain Kani proofs in `tests/proofs_v16.rs` | 244 |
+| Plain Kani proofs in `tests/proofs_v16.rs` | 241 |
 | Plain Kani proofs in `tests/proofs_v16_arithmetic.rs` | 11 |
 | Plain Kani proofs in `src/v16_proofs.rs` | 38 |
 | Function-contract proofs in `src/v16_proofs.rs` | 51 |
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -743,6 +743,24 @@ impl V16Core {
     }
 
     #[inline]
+    fn negative_pnl_accrual_step(attributed: u128, amount: u128) -> V16Result<u128> {
+        attributed
+            .checked_add(amount)
+            .ok_or(V16Error::ArithmeticOverflow)
+    }
+
+    #[inline]
+    fn negative_pnl_reduction_step(attributed: u128, reduction: u128) -> (u128, u128) {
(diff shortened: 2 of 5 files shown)


aeyakovenko:
Added commit ab111c7e with inductive provenance coverage over the production negative-PnL ledger.

New obligations:
- Full-width accrual theorem: selected-domain debt increases exactly, unrelated debt is framed, and aggregate attribution increases exactly.
- Full-width principal theorem: two-domain debt falls by exactly principal paid, no domain increases, and boundary-crossing/full-clear cases are covered.
- Full-width targeted-cure theorem: only the selected domain falls and aggregate debt falls by exactly the cure.
- Sparse binding theorem: production preflight predicts the exact domain/remainder left by production principal mutation, across both sparse orders and compaction.
- Sparse targeted binding theorem: production compaction cannot redirect a domain-specific cure into unrelated debt.

All 30 cover goals are reachable. Final Kani times were 0.95s, 3.88s, 0.71s, 78.06s, and 98.38s, each below 300s.

Mutation evidence:
- Dropped accrual, inert reduction, and inert targeted-cure kernels each fail their full-width theorem.
- Preflight bypass, generic-mutator bypass, and targeted-mutator bypass each fail the corresponding sparse binding theorem.
- Bypassing production accrual causes 8 existing end-to-end v16 regressions to fail.

Verification: `cargo test --all-targets --features fuzz --quiet` passes all 160 tests; formatting and diff checks pass. No spec, layout, or public runtime API changes.

aeyakovenko:

PR #149, opened 2026-07-22
[REAL DoS] Prepare lapsed backing before resolved settlement

aeyakovenko:
## Finding

A finite-lived source backing bucket can lapse while an account retains source-attributed positive PnL and a pending K/F loss. After a valid market resolution, `close_resolved_account_not_atomic` previously ran K/F settlement before expiry preparation. Source-claim burn then consulted the stale Fresh bucket and returned `Stale` forever, stranding the portfolio capital.

## Fix

Commit at most one lapsed-domain preparation before K/F settlement. A persisted lien is released first; a later call expires the bucket. Each successful call decreases a bounded terminal rank and introduces no authority, withdrawal, destination, amount, or custody surface.

## TDD

- RED `b4b087025bd708d012e46bbf8ddd75ff995734ff`: focused pending-K-loss state fails with `Stale`.
- GREEN `e3ea7251ed75f0111ba3cf5a4b87961f6c7cca63`: the first call commits expiry without moving value; bounded continuation closes the account.
- Full `cargo test`: 131 tests passed, plus doc tests.

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -16200,6 +16200,51 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
         Ok(converted)
     }
 
+    /// Commit at most one mutation which makes a lapsed source domain safe for
+    /// resolved settlement. K/F settlement can reduce positive PnL and burn its
+    /// source claim, so this must run before those side effects consult a stale
+    /// Fresh bucket.
+    fn prepare_one_lapsed_source_domain_for_resolved_close_not_atomic(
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #149 (e57296cd73f3f6f7233f2efab223be6b300c6084) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #150, opened 2026-07-23
[REAL DoS] Expire prospective K/F sources before resolved settlement

aeyakovenko:
## Finding

Resolved close can permanently roll back when a leg has positive prospective K/F settlement against a Fresh backing domain that lapsed before the account ever recorded that source domain.

An exact-parent public LiteSVM lifecycle in the stacked wrapper PR creates this state with ordinary deposits, a trade, authenticated mark moves, permissionless cranks, and market resolution. Alternating honest `CloseResolved` calls do not progress: K/F settlement tries to create the missing positive-PnL claim, account valuation rejects its lapsed backing, and SVM rollback preserves the blocker.

This is distinct from #149: #149 prepares lapsed source domains already present in the account. This case concerns a source domain that is only discoverable from the pending K/F settlement.

## Fix

- Before resolved K/F settlement, scan the bounded active-leg roster for a positive prospective K/F source whose Fresh backing has lapsed.
- Expire one such source domain and return `ProgressOnly`, so every successful call strictly removes one blocker.
- Skip the preflight and duplicate validation for flat portfolios; this preserves the existing max-domain close CU bound.
- Expose the production predicate to Kani and prove that the preflight strictly decreases the count of prospective lapsed sources.

## Verification

- Engine production regression: `v16_resolved_close_expires_lapsed_prospective_kf_sources`
- Kani: `proof_v16_resolved_prospective_source_expiry_strictly_decreases_rank` passes all assertions and 3/3 covers
- Engine suite: 168 tests pass
- Stacked wrapper exact-parent RED becomes green
- New public 14-leg full-roster CU test stays below 1,000,000 CU
- Existing 28-domain resolved-close CU regressions remain green
- Full wrapper matrix on the final pin: 5 unit + 575 LiteSVM/CU tests pass

Stacked on #149 because the public reproduction requires that earlier resolved-settlement ordering fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Source-checkable counts in this checkout:
 
 | Class | Count |
 | --- | ---: |
-| Plain Kani proofs in `tests/proofs_v16.rs` | 249 |
+| Plain Kani proofs in `tests/proofs_v16.rs` | 255 |
 | Plain Kani proofs in `tests/proofs_v16_arithmetic.rs` | 11 |
 | Plain Kani proofs in `src/v16_proofs.rs` | 38 |
 | Function-contract proofs in `src/v16_proofs.rs` | 51 |
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Source-checkable counts in this checkout:
 
 | Class | Count |
 | --- | ---: |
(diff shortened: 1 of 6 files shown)


aeyakovenko:

PR #151, opened 2026-07-23
[REAL DoS] Expire prospective loss backing before resolved settlement

aeyakovenko:
## What changed

- Extend resolved-close prospective-domain preflight to inspect both K/F polarities.
- For a pending negative K/F delta, select the leg's same-side capital-backing domain and expire one lapsed `Fresh` bucket before settlement.
- Add a production-route regression and a Kani rank-decrease proof for the new branch.

## Root cause and impact

This is stacked on #150. That fix preflights a pending positive K/F source claim, but a pending negative K/F delta can also mutate source-domain state: settlement reserves newly negative PnL as same-side capital backing.

If that same-side bucket is still marked `Fresh` after its expiry, settlement attempts to add backing with a new expiry and returns `LockActive`. In a resolved market the account cannot trade, rebalance, or use live crank routes, so every bounded public `CloseResolved` retry rolls back and the portfolio's capital remains locked. A public two-market LiteSVM lifecycle on the exact parent reproduces this without state injection; other portfolios reach their terminal paths while the victim repeatedly returns `EngineLockActive`.

The fix performs one bounded, value-neutral expiry step and returns progress. A later close settles the loss normally.

## Validation

- `cargo fmt --check`
- `cargo test --all-targets` (50 unit, 11 backing, 2 resolved-insolvent, 78 spec tests)
- `cargo kani --tests --features fuzz --harness proof_v16_resolved_prospective_loss_backing_expiry_strictly_decreases_rank` (0/16 failed; 3/3 covers)
- Stacked wrapper LiteSVM RED/GREEN reproduction and 14-leg max-shape CU test

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -682,6 +682,25 @@ impl V16Core {
             )
     }
 
+    #[inline]
+    fn resolved_prospective_loss_backing_requires_expiry(
+        pending_kf_delta: i128,
+        bucket_status: BackingBucketStatusV16,
+        bucket_expiry_slot: u64,
+        current_slot: u64,
+    ) -> bool {
+        pending_kf_delta < 0
+            && matches!(
+                Self::resolved_source_close_phase(
+                    0,
+                    bucket_status,
+                    bucket_expiry_slot,
+                    current_slot,
+                ),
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #152, opened 2026-07-23
[REAL DoS] Reconcile foreign-expired source liens during resolved close

aeyakovenko:
## Impact

A publicly reachable resolved market can permanently strand a source-backed winner. Account A creates a legitimate source-credit IM lien through ordinary trades. A lien-free account B closes first after authenticated expiry and moves the shared aggregate bucket to Impaired. A still records its account-local lien as live, so its terminal release debits an aggregate live-lien counter that is already zero and returns CounterUnderflow. SVM rollback restores the same state; direct CloseResolved and the sole public crank repeat the failure.

The first version of this fix restored only the initial live-to-impaired relabel. PR135's full public lifecycle found a second CounterUnderflow: that relabel placed expired provider backing into the generic impaired-effective field, whose burn path correctly assumes impaired insurance and tries to release an insurance reservation that does not exist.

## Fix

Before resolved settlement releases a source lien:

1. Charge utilization only through the authenticated bucket expiry.
2. While the account still carries its exact counterparty amount, retire exactly that amount from the aggregate provider-impaired bucket/source ledgers.
3. Convert the account claim face from live to zero-credit impaired face.
4. Do not add expired provider principal to the insurance-backed impaired-effective reserve.
5. Normalize the shared bucket from Impaired to Expired only when the final provider label is retired.

This is order-independent across accounts: one account cannot consume a sibling's provider label, and all insurance reservation fields are framed.

## TDD evidence

- RED `dfd7eed4`: public engine lifecycle reproduces the foreign-expiry live-lien mismatch.
- Partial green `1e0d952e`: restores the first bounded relabel.
- RED `c30540df`: strengthens the regression through complete terminal settlement and exposes the downstream insurance CounterUnderflow.
- GREEN `0a23b5f5`: preserves provider/insurance provenance and closes the full lifecycle.
- The runtime regression uses two independently liened winners sharing one expired bucket. The first leaves the sibling label intact; all six funded accounts terminate and the final aggregate impaired provider label reaches zero.

PR135 independently reaches the state through public SBF instructions. On `1e0d952e`, the first relabel succeeds and every later direct CloseResolved/public-crank retry returns Custom(25) with exact SVM rollback.

## Verification

- `cargo test --all-targets --features fuzz`: 183 tests pass.
- `proof_v16_resolved_foreign_expiry_lien_impairment_is_exact_relabel`: 422 checks, 3/3 covers, PASS.
- `proof_v16_resolved_foreign_expiry_retires_exact_provider_label_only`: 108 checks, 3/3 covers, PASS.
- `proof_v16_resolved_foreign_expiry_provider_retirement_acceptance_is_exact`: 36 checks, 4/4 covers, PASS.
- Source proof roster: 270 plain v16 Kani harnesses.

Stacked on #153 because that fix removes the earlier impaired-domain prospective-loss blocker and exposes this terminal rank failure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Source-checkable counts in this checkout:
 
 | Class | Count |
 | --- | ---: |
-| Plain Kani proofs in `tests/proofs_v16.rs` | 267 |
+| Plain Kani proofs in `tests/proofs_v16.rs` | 268 |
 | Plain Kani proofs in `tests/proofs_v16_arithmetic.rs` | 11 |
 | Plain Kani proofs in `src/v16_proofs.rs` | 38 |
 | Function-contract proofs in `src/v16_proofs.rs` | 51 |
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Source-checkable counts in this checkout:
 
 | Class | Count |
 | --- | ---: |
-| Plain Kani proofs in `tests/proofs_v16.rs` | 268 |
+| Plain Kani proofs in `tests/proofs_v16.rs` | 270 |
 | Plain Kani proofs in `tests/proofs_v16_arithmetic.rs` | 11 |
 | Plain Kani proofs in `src/v16_proofs.rs` | 38 |
 | Function-contract proofs in `src/v16_proofs.rs` | 51 |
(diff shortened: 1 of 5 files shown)


aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #152 (0a23b5f5fc85ddb0223089c66c29cbf1600be62b) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #153, opened 2026-07-23
[REAL DoS] Route prospective loss after source impairment

aeyakovenko:
## Impact

A publicly reachable resolved market can strand an ordinary losing portfolio. If another account first expires a shared source bucket containing a valid lien, the bucket becomes Impaired. A portfolio with an unsettled adverse K/F delta then tries to crystallize capital-backed loss as new Fresh provider principal. The impaired bucket rejects that transition with LockActive, so every CloseResolved retry rolls back and the portfolio cannot wind down.

## Fix

In Resolved mode only, route newly crystallized loss for an already-Impaired source domain directly into terminal junior residual. Keep the impaired bucket and recoverable provider-principal totals unchanged. If the payout snapshot exists, credit the exact atoms through the post-snapshot payout-ledger kernel from #147.

## TDD evidence

- RED commit 6e0dc19c reproduces LockActive using public engine lifecycle operations on the exact #147 head.
- GREEN commit cc83730b restores bounded CloseResolved progress.
- PR135 independently reaches the state through ordinary wrapper instructions; 16 alternating CloseResolved/crank attempts return Custom(21) with exact rollback on the vulnerable pin.

## Verification

- cargo test --all-targets --features fuzz: 182 tests pass.
- proof_v16_resolved_impaired_capital_backed_loss_routes_to_residual: 6,112 checks, 4/4 nonvacuity covers, PASS.
- The proof frames vault value, capital/c_tot, PnL, residual, provider buckets, fresh totals, crystallized-loss counters, and both snapshot states.

Stacked on #147 because this fix reuses its post-snapshot residual-credit kernel.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Source-checkable counts in this checkout:
 
 | Class | Count |
 | --- | ---: |
-| Plain Kani proofs in `tests/proofs_v16.rs` | 266 |
+| Plain Kani proofs in `tests/proofs_v16.rs` | 267 |
 | Plain Kani proofs in `tests/proofs_v16_arithmetic.rs` | 11 |
 | Plain Kani proofs in `src/v16_proofs.rs` | 38 |
 | Function-contract proofs in `src/v16_proofs.rs` | 51 |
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #153 (cc83730beffd78ff3d7a26f23363a4ca254eef34) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #154, opened 2026-07-23
[REAL DoS] Align cross-domain source support rounding

aeyakovenko:
## Impact
A public two-asset lifecycle can leave a live portfolio permanently unable to settle a fair mark reversal. The support estimator floors fractional source support only after summing domains, while the consumer floors each domain independently. The estimator can therefore request one atom that no source domain can supply; both permissionless auto-crank and owner rebalance return `LockActive` forever under honest later-slot retries.

## Reproduction
The first commit is an exact-parent engine RED: two source domains each provide 100 whole atoms plus fractional support, the estimator advertises 201 atoms, and the mutator can consume only 200. No state injection or dishonest oracle is required.

## Fix
Use one per-domain `source_credit_consumable_atoms` calculation in both support estimators and the mutating consumer. This preserves conservative per-domain rounding and available-backing caps.

## Verification
- `cargo test --all-targets --features fuzz --quiet` (159 tests)
- focused Kani estimator/consumer compatibility proof: 458 checks, 2/2 covers, 0 failures
- exact wrapper LiteSVM RED/GREEN is being published separately against this commit

Exact parent: `143e68c4917ed0400a27b952f036a5677047cd84`.

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -716,6 +716,31 @@ impl V16Core {
         Ok((required_face_num, required_backing_num))
     }
 
+    #[cfg_attr(all(kani, feature = "contracts"), kani::ensures(|result: &V16Result<u128>| match result {
+        Ok(value) => *value <= unliened_claim_num / BOUND_SCALE
+            && *value <= available_backing_num / BOUND_SCALE,
+        Err(_) => true,
+    }))]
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #155, opened 2026-07-23
[REAL DoS] Cap B settlement by loss atoms

aeyakovenko:
## Finding

`account_b_settlement_chunk_from_leg` derived an atom-bounded B-index delta, then also capped that delta directly by `public_b_chunk_atoms`. The configuration and endpoint budgets are collateral atoms, not B-index units.

A valid public market lifecycle at the maximum supported quote scale produces a B target of `107486458947473684210526315` with survivor loss weight `19000000`. The total realizable loss is about `2.042e12` atoms, safely below the default `1e16` atom budget, but the old implementation advanced B by only `1e16` per successful close. Terminal settlement therefore required more than 10.7 billion transactions and stranded survivor funds despite an honest cranker.

## Fix

- Treat both configured and endpoint limits as loss-atom budgets.
- Take their minimum before converting once through `loss_weight`.
- Remove the dimensionally invalid direct B-index cap.
- Rename internal budget parameters to make the unit explicit.

The first commit is the exact-parent RED regression. The second commit is the narrow fix plus a focused Kani regression.

## Verification

- `cargo test --all-targets --features fuzz --quiet`
- `cargo kani --features fuzz --harness proof_v16_b_settlement_atom_budget_clears_public_scale_gap`
  - 2,843 checks, 0 failures

## Dependency

This is intentionally stacked on #115. Exact `master` reaches the independent ADL terminal-close blocker from #115 first; applying only #115 exposes this B-settlement unit bug through the public wrapper lifecycle.

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -10866,7 +10866,7 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
         &mut self,
         account: &mut PortfolioV16ViewMut<'_>,
         price_override: Option<(usize, u64)>,
-        b_delta_budget: u128,
+        b_loss_atom_budget: u128,
         allow_b_chunk: bool,
     ) -> V16Result<AccountRefreshCertOutcomeV16> {
         self.validate_account_scalar_preflight(&account.as_view())?;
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #157, opened 2026-07-24
[BLOCKER LoF] Recredit claim-free terminal overlap

aeyakovenko:
## Finding

A public liquidation can spend directional insurance for a bankrupt account while the resolved winner later consumes counterparty backing for the same loss. After every portfolio exits, the provider is short by the overlapping amount, the residual is ownerless, and terminal close is blocked.

Tracked by aeyakovenko/percolator-prog#289.

## Fix

- Add an asset-local terminal sweep gated on resolution, zero materialized portfolios, and zero outstanding trader claim/blocker counters.
- Reclassify residual to the paired insurance domain only up to all three independent caps: claim-free residual, historical paired-domain insurance spend, and outstanding source-provider receivable.
- Preserve vault custody while restoring insurance and reducing the paired domain spend.
- Keep live and per-winner source conversion unchanged, so close order cannot reclassify support still owed to another winner.
- Keep the API asset-local so one recovery call touches exactly two source domains instead of scanning a maximal market slab.

## Verification

- Full engine matrix: 50 unit, 11 backing fuzz, 15 reference, 2 insolvent, 9 rounding, 1 rollback fuzz, and 72 spec tests green.
- Deterministic terminal-overlap, mixed exact-support, and live-conversion regressions green; the mixed-order case proves ordinary winner support is not recredited.
- Kani pure cap proof: 0/25 failed, all 3 cap branches covered.
- Kani whole-mutation proof: 0/3579 failed, required nonzero mutation covered.
- Exact-parent public LiteSVM RED/GREEN is in the linked wrapper PR; the final provider recovery consumes 126,863 CU on a 5,834-slot market.

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -717,6 +717,17 @@ impl V16Core {
     }
 
     #[inline]
+    fn terminal_claim_free_overlap_recredit(
+        provider_receivable_atoms: u128,
+        paired_domain_insurance_spent: u128,
+        claim_free_residual_remaining: u128,
+    ) -> u128 {
+        provider_receivable_atoms
+            .min(paired_domain_insurance_spent)
+            .min(claim_free_residual_remaining)
+    }
+
+    #[inline]
     fn validate_bound_num_atom_aligned(bound_num: u128) -> V16Result<()> {
         if bound_num == 0 {
             return Ok(());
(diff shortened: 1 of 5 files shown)


aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #157 (76a86f483dd1dd3d3fffa51bea1bba67a8168411) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #159, opened 2026-07-27
[BLOCKER LoF] Settle authenticated mark loss before fee sync

aeyakovenko:
## Finding

A permissionless caller can front-run an already-authenticated adverse mark with `SyncMaintenanceFee`. Before the fix, fee sync settles only realized K/F/PnL and ignores target/effective mark lag.

The public wrapper LiteSVM exploit produced:

- mark first: cranker reward 50,000; victim payout 0; independent winner payout 50,000
- fee sync first: cranker reward 100,000; victim payout 0; independent winner payout 0

The attacker needs no authority. The authenticated mark update is already committed; the attacker invokes only the permissionless fee-sync route and withdraws its reward.

## Fix

For a live non-flat account, reject fee sync with `LockActive` while any exposed price-managed asset has target/effective lag. A public crank can apply the committed mark, after which fee sync succeeds. This does not block trading or exits.

The guard is a production kernel with an exact Kani contract over all Boolean combinations. The existing hidden-K loss test now stages the raw target before accrual, matching production ordering and retaining coverage for the fully-applied-mark path.

## Verification

- Exact pre-fix native RED: sync returned `Ok(20)` instead of `Err(LockActive)`
- Focused native GREEN
- Kani guard proof: 0 failures, 2/2 covers
- `cargo test --all-targets --features fuzz -- --test-threads=1`: 159 passed
- `git diff --check`

A paired wrapper PR will add the public LiteSVM regression and guard wrapper-profile lag until authenticated targets are staged into the engine at submission time.

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -12306,13 +12306,36 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
     fn account_has_target_effective_lag(&self, account: &PortfolioV16View<'_>) -> V16Result<bool> {
         let mut slot = 0usize;
         while slot < V16_MAX_PORTFOLIO_ASSETS_N {
-            let leg = account.header.legs[slot].try_to_runtime()?;
-            if leg.active && self.asset_has_target_effective_lag(leg.asset_index as usize)? {
+            let leg = &account.header.legs[slot];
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #160, opened 2026-07-27
[BLOCKER DoS] Bound max-source PnL conversion

aeyakovenko:
## Finding

A public LP can accumulate all 32 sparse source domains through ordinary matcher fills and honest mark movement. Once flat, `ConvertReleasedPnl` attempts to realize every source claim in one engine call. At the legal maximum shape it consumes exactly 1,400,000 CU and rolls back, so the LP cannot realize or withdraw material source-backed profit while the healthy market remains Live.

The paired wrapper regression uses only public LiteSVM instructions. The LP authorizes the matcher once; an unrelated taker alone signs the subsequent fills. The exact-parent RED leaves the victim with 32 occupied domains and 32,000 atoms of backed PnL, then the first conversion aborts at the SVM ceiling.

## Fix

Add a bounded conversion primitive that realizes at most one supported source domain per call. Each successful call:

- consumes backing and burns claim face in the same domain
- removes at most one sparse source entry
- decreases positive PnL and increases capital by the proven conversion amounts
- preserves the existing caller cap in the wrapper and the existing all-at-once engine API for internal callers

A health refresh remains required between steps because conversion invalidates the account certificate; this gives a finite public `convert -> auto-crank` sequence without weakening risk validation.

## Dependency

Stacked on #140 so the bounded path reuses its domain-local claim-burn invariant instead of duplicating or conflicting with that fix.

## Verification

- `cargo test --all-targets -- --test-threads=1`: 50 unit/math + 9 backing fuzz + 2 resolved-insolvency fuzz + 70 V16 spec tests passed
- `cargo test --features fuzz --all-targets -- --test-threads=1`: all runtime, conformance, fuzz, and 74 feature-gated V16 spec tests passed
- focused two-domain rank regression passed

Paired public wrapper PR follows after exact fixed-head LiteSVM/CU verification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ Source-checkable counts in this checkout:
 | Plain Kani proofs in `src/v16_proofs.rs` | 38 |
 | Function-contract proofs in `src/v16_proofs.rs` | 51 |
 | Production `kernel_*` helpers in `src/v16.rs` | 17 |
-| Public `*_not_atomic` engine APIs in `src/v16.rs` | 55 |
+| Public `*_not_atomic` engine APIs in `src/v16.rs` | 56 |
 
 Spot-check the inventory directly:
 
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #161, opened 2026-07-27
[BLOCKER DoS] Bound resolved close work per call

aeyakovenko:
## What changed

- clear at most one active leg per `close_resolved_account_not_atomic` call
- realize at most one source-claim domain when multiple domains remain
- return `ProgressOnly` after each bounded step and validate the committed state
- preserve the existing one-domain realization semantics
- add a production-API engine regression that opens two legs, resolves, and proves each close call makes value-neutral progress before payout

## Root cause

Resolved close detached every active leg and realized every source domain in one invocation. A publicly constructed maximum-shape portfolio can therefore exceed the SVM compute limit after permissionless resolution. SVM rollback restores the same state, so retrying the only terminal close route cannot make progress and funds remain locked.

## Impact

This turns terminal close into a bounded rank-decreasing operation. Repeated permissionless calls remove one leg or one source domain at a time and eventually complete without requiring a privileged recovery path.

## Validation

- `cargo test --all-targets` (132 passed)
- `cargo test --test backing_double_claim_fuzz` (9 passed)

Stacked on the one-source-domain conversion primitive in PR #160; the public LiteSVM red/green regression is being added in the corresponding wrapper PR.

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -8429,6 +8429,22 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
         Ok(Self::account_source_claim_bound_sum_num(account)? != 0)
     }
 
+    fn account_source_claim_domain_count(account: &PortfolioV16View<'_>) -> usize {
+        let mut count = 0usize;
+        let mut slot = 0usize;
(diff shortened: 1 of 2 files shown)


aeyakovenko:

PR #162, opened 2026-07-27
[BLOCKER LoF] Track stale K/F settlement cohorts

aeyakovenko:
## What changed

- Add a checked monotonic K/F settlement generation to each asset side and snapshot it on every leg.
- Count the positions that predate each K/F change, consume that work exactly once on settlement, and preserve the count across later accrual segments.
- Keep a `ResetPending` side's K/F endpoint and generation frozen while the other side and market time continue to accrue.
- Treat source-backed positive PnL as non-favorable while its counterparty side has stale K/F work, and keep HMax active during the one-step negative-PnL interval before bankruptcy booking.
- Preserve engine trading during cohort settlement. Fully signed position transfer and risk reduction remain available; wrappers exposing unsigned counterparties must use the counters at that authorization boundary.
- Bump the persisted layout discriminator from 17 to 18 for the new asset and leg fields.

## Root cause

K/F staleness was inferred only from numeric endpoint equality. A price or funding round trip could therefore return to the old K/F value and erase evidence that accounts still belonged to an earlier settlement cohort. More importantly, the engine had no bounded summary of losing accounts that had not yet settled. A winner could obtain source-backed PnL before those losses became bankruptcy residual, leaving an unsigned wrapper LP able to inherit the later social-loss B charge as fresh exposure.

## Security effect

The new generation survives endpoint round trips and cannot be consumed twice. Favorable source-credit use remains locked until counterparties settle and any resulting negative PnL is closed. The engine does not impose a market-wide admission pause: a high-stress fully signed novation is covered and remains successful. The companion wrapper PR consumes the new side counters to reject only fresh unsigned LP exposure to historical loss.

## Verification

- `cargo test --all-targets --features fuzz`: pass, including 75/75 v16 spec tests.
- Exact K/F generation round-trip and one-consumer-per-leg regression: pass.
- High-stress signed winner novation during stale bankruptcy cohort: pass.
- Kani risk-admission theorem: 0/2960 failed, 6/6 covers.
- Kani stale-source HLock theorem: 0/3880 failed, 5/5 covers.
- Kani global negative-PnL HLock theorem: 0/3218 failed, 6/6 covers.
- Kani accrual/reset cohort theorems: 0/3349 and 0/3177 failed.
- Changed function contracts: attach 0/1504, resize 0/1390, clear 0/1469, B-snapshot advance 0/385 failed.
- Plain and function-contract Kani code generation: pass.

The public LiteSVM red/green exploit and unsigned single/batch wrapper enforcement are in the companion program PR.


aeyakovenko:

PR #163, opened 2026-07-28
[BLOCKER DoS] Release flat source liens before PnL conversion

aeyakovenko:
## What changed

- Compile the existing source-lien release kernels into production.
- Keep full conversion behavior for direct engine callers, while `convert_released_pnl_to_capital_step_not_atomic` releases at most one source domain and converts at most one source domain per call.
- Release only after the favorable-action preflight and only when capital without positive source credit satisfies current initial margin.
- Clear the completed lien's live fee cursor/counter so its sparse source slot can dematerialize.
- Extend the production lifecycle test through risk increase, flattening, lien release, and PnL realization.

## Root cause

The no-DoS proof surface included `release_account_source_credit_liens_if_unneeded_not_atomic`, but that transition and its lower release kernels were behind `cfg(kani|fuzz)`. A normal account could use source-backed positive PnL as initial margin, close every leg, and retain a now-unneeded lien. Production conversion could consume only unliened claims, so it returned `LockActive` forever even though the account was flat and healthy.

## Stack / merge prerequisites

This PR is stacked on #160 because the wrapper uses its bounded one-source conversion API. Releasing every lien before that step would reintroduce max-shape CU work; this version preserves the one-domain rank decrease.

The final rebased engine lineage must also carry `14c2919d` (companion wrapper #306), which aligns aggregate source support with whole atoms consumable by individual domains. Without that independent fix, any bounded per-domain converter can inherit the already-known cross-domain rounding stall.

## Validation

- `cargo fmt --check`
- `cargo test --all-targets`: 131/131 passed
- Companion public LiteSVM regression passes with production SBF
- Combined 32-source public regression completes every conversion; worst call is 1,321,211 CU, 78,789 below the transaction limit
- Exact parent RED: `ConvertReleasedPnl` returned `LockActive` after both user legs closed successfully

Companion wrapper PR: https://github.com/aeyakovenko/percolator-prog/pull/364

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -2407,7 +2407,6 @@ impl V16Core {
         Ok((bucket, source))
     }
 
-    #[cfg(any(kani, feature = "fuzz"))]
     // Contract layer: lien release is the un-pledge relabel — valid liened
     // returns to fresh unliened atom-for-atom, fresh_reserved and all stock
     // quantities untouched; zero-amount is the identity.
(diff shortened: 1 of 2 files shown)


aeyakovenko:

PR #164, opened 2026-07-29
[BLOCKER LoF] Carry fractional oracle caps across max-dt cranks

aeyakovenko:
## Finding

The zero-delta anti-pinning fix preserves elapsed time only until a representable move exists. At low prices, however, even the configured maximum accrual segment can round below one price atom forever. With price 100, target 1, a 24 bps/slot cap, and max dt 20, the effective price deterministically stalls at 20.

A public LiteSVM wrapper repro on top of engine PR #128 uses two independent owners and an honest AuthMark push. Permissionless cranks reach 20, then remain `NonProgress` even after arbitrarily later calls. Permissionless stale resolution freezes 20 and pays 999,920 / 1,000,080 instead of target-valued 999,901 / 1,000,099, transferring 19 atoms from the short to the long without an admin or oracle compromise.

## Fix

Carry the fractional percentage-cap numerator across bounded accrual segments. A sub-atom segment persists a remainder below 10,000; once accumulated budget represents a whole target-directed atom, consuming the segment without moving still rejects. Target convergence and direction reversals reset the remainder, while continuously advancing same-direction reports preserve it. This preserves the exact cumulative cap instead of weakening it with a one-atom ceil or allowing a moving honest feed to discard every fraction.

The carry reuses `retired_slot` only while an asset is Active/DrainOnly; retirement overwrites it with the actual retirement slot, so the persisted layout and account sizes do not change.

## Verification

- `cargo test --all-targets -- --test-threads=1`: 133 tests green
- Fixed-target finite-progress regression reaches price 1 and observes fraction-only segments
- Moving same-direction targets retain enough carry to advance; an opposite-direction target resets it
- Existing zero-delta anti-pinning regression remains green

Stacked on engine PR #128.


aeyakovenko:

PR #165, opened 2026-07-29
[BLOCKER DoS] Resolve terminal recovery through auto-crank

aeyakovenko:
## Impact

A normally initialized market can enter a permanent global `Recovery` sink after a legitimate asset shutdown. A portfolio owner can publicly start a chunked `ForfeitRecoveryLeg`, leave a nonzero residual until its close deadline expires, and let an unsigned `PermissionlessCrank` commit `Recovery`. Unrelated flat users then lose every bounded principal-exit path.

The primary LiteSVM reproduction uses only public wrapper instructions and normal accounts. It does not mutate protocol state. It is RED against wrapper `36fa587e` with engine `143e68c4917ed0400a27b952f036a5677047cd84`: the expiry crank leaves mode `Recovery`, and the unrelated user's funds are globally locked.

## Root cause

`AutoCrankPlanV16::DeclareRecovery` committed `MarketModeV16::Recovery`, while `resolve_market_not_atomic` rejected that same mode. The actionable-summary builder also classified a persisted `Recovery` state as having no continuation, so upgrading the program could not rescue an already committed state through the sole public auto-crank.

## Fix

- Complete `DeclareRecovery -> Resolved` in the same auto-crank call.
- Permit the value-neutral engine resolution transition from `Recovery` while retaining its audit reason.
- Classify a persisted `Recovery` state as `recovery_eligible`, preserving the recorded reason and routing it through the same auto-crank.
- Add runtime and Kani coverage for fresh expiry and pre-upgrade persisted states.

## Verification

- `cargo test --all-targets -- --test-threads=1`: 130/130 passed.
- `proof_v16_public_resolve_market_is_value_neutral_and_clears_loss_stale`: 4,543 checks passed, 3/3 covers.
- `proof_v16_committed_recovery_auto_crank_resolves_value_neutrally`: 14,839 checks passed over symbolic capital, insurance, and surplus.
- Associated wrapper branch: SBF build passed; 567/567 LiteSVM/CU tests plus wrapper unit tests passed.
- Exact parent/head public TDD: parent commits global `Recovery`; fixed head reaches `Resolved` and pays an unrelated flat owner's 100 real SPL tokens through `CloseResolved`.

## Known dependency

This PR restores the global transition and the unrelated flat-user exit. Full wind-down of the loss-side cohort under a small valid B budget additionally depends on #155: without that independent unit fix, `public_b_chunk_atoms` is incorrectly reused as a B-index delta cap and terminal settlement can require billions of calls. This PR intentionally does not duplicate #155.

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -11582,7 +11582,7 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
     ///   pending_close    — Live, a close-progress ledger is active
     ///   expired_close    — Live, that ledger is past its max-close slot
     ///   liquidatable     — Live, current cert with nonzero certified liq deficit
-    ///   recovery_eligible— Resolved, unattributed-insolvent negative-PnL recovery
+    ///   recovery_eligible— Recovery, terminal resolution continuation
     ///   resolved_winner  — Resolved, positive PnL, resolved payout ready
     /// Assembled via the proven actionable_summary_from_signals kernel. Live-only
     /// flags need cert currentness only where their entrypoint does (liquidate),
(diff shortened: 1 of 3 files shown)


aeyakovenko:

PR #166, opened 2026-07-29
[BLOCKER DoS] Keep winner-first Recovery forfeits live

aeyakovenko:
## Impact

After a normal secondary-asset shutdown, the winning owner can publicly forfeit first. The bankrupt owner's subsequent `ForfeitRecoveryLeg` used to return `RecoveryRequired`; wrapper/SVM rollback erased both its declaration and close-ledger progress. Retry, unilateral reduction, pair force-close, and auto crank had no successful continuation, while an honestly refreshed base oracle kept whole-market stale resolution unavailable. A real provider-backing lien created through public cross-margin risk also remained stranded.

The paired wrapper LiteSVM reproduction uses only ordinary deposits, provider backing, trades, bounded authenticated marks, shutdown, owner forfeits, and permissionless crank instructions. It does not inject protocol state.

## Root cause

The winner could detach after positive PnL and its source claim had already been materialized. The old forfeit path removed the selected leg and its social-loss weight but preserved that materialized claim and valid source lien. This contradicted the dead-leg rule in `spec.md`: an owner-forfeited leg's positive PnL must be valued at zero unless source backing is consumed. With no remaining recipient weight and a stale source claim, the bankrupt counterparty could not settle its residual.

## Fix

The selected-leg forfeit now:

- waives that leg's already materialized positive PnL and matching source claim;
- releases valid source liens back to backing instead of consuming provider principal;
- isolates the selected source slot before canonical claim/PnL burn, preserving unrelated source domains;
- permits explicit terminal loss only for a claim-free local Recovery domain with no provider receivable or backing/insurance lien.

The accounting remains conservative: any selected-domain attribution residue is also forfeited, and unrelated PnL, claims, backing, insurance, and liens remain unchanged.

## TDD

- `v16_claim_free_recovery_domain_survives_winner_first_forfeit`: the winner detaches and the loser then clears a claim-free terminal residual.
- `v16_recovery_forfeit_waives_materialized_winner_claim_before_detach`: refreshes/materializes the winner first, proves its selected claim is waived, and proves the loser remains live.
- `v16_recovery_forfeit_preserves_unrelated_source_domain_claim`: proves the selected-leg waiver cannot burn another asset's claim or PnL.
- Wrapper PR aeyakovenko/percolator-prog#371 supplies the end-to-end public LiteSVM red/green reproduction, including a provider-backing lien created by public cross-margin trading.

Against exact parent `143e68c4917ed0400a27b952f036a5677047cd84`, the wrapper loser forfeit fails with `Custom(23)` at 77,366 CU and 22.6k atoms of provider backing remain liened. With this commit, all three portfolios close and the provider withdraws the full 100k backing while the unrelated base market remains live.

## Verification

- `cargo test --all-targets -- --test-threads=1`: 132 passed
- `contract_check_kernel_terminal_explicit_loss_allowed`: 211 checks, 0 failures
- `contract_check_kernel_bresidual_step`: 147 checks, 0 failures
- Wrapper full suite: 566/566 LiteSVM tests pass
- `git diff --check`

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -15713,6 +15713,76 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
         ))
     }
 
+    fn forfeit_materialized_positive_pnl_for_leg(
+        &mut self,
+        account: &mut PortfolioV16ViewMut<'_>,
+        asset_index: usize,
+        leg_side: SideV16,
+    ) -> V16Result<u128> {
+        let source_domain = self.insurance_domain_index(asset_index, opposite_side(leg_side))?;
+        let Some(mut source_slot) = account.source_domain_slot(source_domain)? else {
+            return Ok(0);
+        };
+        let source_claim_bound_num = account.header.source_domains[source_slot]
+            .source_claim_bound_num
+            .get();
+        if source_claim_bound_num == 0 {
+            return Ok(0);
+        }
+        V16Core::validate_bound_num_atom_aligned(source_claim_bound_num)?;
+
+        // A dead-leg forfeit values this leg's positive PnL at zero. Release any
+        // valid source-credit lien first so forfeiting the face returns backing
+        // instead of consuming it. Impaired claim fields are burned by the
+        // canonical source-claim burn below.
+        if account.header.source_domains[source_slot]
+            .source_claim_liened_num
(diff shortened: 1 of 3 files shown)


aeyakovenko:

PR #167, opened 2026-07-29
[BLOCKER LoF] Preserve prior claims across Recovery forfeits

aeyakovenko:
## Impact

A public, normally initialized market can erase a user's settled source-backed PnL from an earlier position episode.

Public LiteSVM sequence:

1. A victim earns and fully closes a profitable short, leaving a 48,000-atom same-domain claim.
2. The victim later reopens the same asset and side at the current mark.
3. The asset enters normal Recovery and the new counterparty forfeits first.
4. Conversion, unilateral reduction, and pair close are unavailable while an unrelated base asset stays honestly current.
5. On exact parent `a9a8588483ac6ecccd5d94bcbc80e116cf5487a8`, the victim's unilateral forfeit burns the prior claim.

The strengthened public case also earns new PnL in episode two and increases risk until its source lien crosses the historical claim floor. The first version of this fix returned `LockActive` there, persistently blocking the victim's only remaining exit.

No account bytes or private engine state are injected by wrapper PR aeyakovenko/percolator-prog#372.

## Fix

Snapshot the source-domain claim bound when a new leg episode attaches. Recovery forfeit burns only the selected episode's suffix above that floor.

When a later valid lien overlaps the historical floor, the same forfeit transition now:

1. releases the domain lien,
2. realizes at most the saved historical floor from that same domain into senior capital,
3. reduces the floor by the exact face burned, and
4. waives only the current episode's remaining suffix.

The release and realization are atomic, so the backing provider cannot withdraw between them. Normal leg clear removes the episode marker, persisted claim burns clamp it, validation binds it to the matching active asset/side, and the layout discriminator is 18.

This PR is intentionally stacked on #166 because it corrects an episode-attribution edge exposed by that Recovery liveness fix.

## Verification

- `cargo test --all-targets -- --test-threads=1`: 133 passed
- `cargo check --all-targets`
- Kani `closure_recovery_forfeit_source_plan_partitions_claim_without_overlap`: 64 checks, 0 failures
- wrapper PR #372 exact-parent RED / fixed-head GREEN
- public overlap case exits in one forfeit, atomically capitalizes exactly 48,000 historical atoms, permits immediate withdrawal of all unconsumed provider backing, and still pays the victim all 53,000 remaining atoms

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -28,7 +28,8 @@ pub const V16_EMPTY_ACTIVE_BITMAP: V16ActiveBitmap = [0; V16_ACTIVE_BITMAP_WORDS
 pub const V16_BACKING_BUCKETS_PER_DOMAIN: usize = 1;
 // Bump whenever the on-chain account/header Pod layout changes (see the
 // PortfolioAccountV16Account size assertion). 17: added funding flow counters.
-pub const V16_LAYOUT_DISCRIMINATOR: u16 = 17;
+// 18: preserve pre-existing same-domain claims across a later leg episode.
+pub const V16_LAYOUT_DISCRIMINATOR: u16 = 18;
 pub const V16_ACCOUNT_VERSION: u16 = 1;
 pub const BACKING_FEE_RATE_DEN_E9: u128 = 1_000_000_000;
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #168, opened 2026-07-29
[BLOCKER LoF] Net Recovery haircuts within position episodes

aeyakovenko:
## Impact

An unprivileged relayer can land a retained, honestly signed `ForfeitRecoveryLeg` transaction after the counterparty changes public market state. On the exact stacked parent, a newly booked `B` haircut from the reopened position consumes an older, fully closed, provider-backed same-domain claim.

The public LiteSVM reproduction creates a 48,000-atom historical claim, reopens the same side, and earns 59,424 atoms in the new episode. The victim signs while its `B` snapshot is current. The underfunded counterparty forfeits first and books an 8,424-atom haircut; replaying the unchanged victim transaction then reduces the old claim from 48,000 to 39,576.

This is classified as `[BLOCKER LoF]`: the victim transaction was safe when signed, and an unprivileged relayer can reorder it within the valid blockhash window after another public transaction changes the haircut target.

## Root cause

Recovery forfeiture discarded the current episode's positive claim suffix before settling that episode's pending `K`, `F`, and `B` deltas. Once the suffix was gone, the later `B` settlement reached below the attach-time claim floor and charged the prior episode.

## Fix

Materialize pending positive `K`/`F` into the selected source domain, promote the selected episode source first, settle bounded `B` while the current-episode suffix still exists, and only then forfeit positive value above the attach-time floor. Partial `B` progress preserves the suffix until settlement is complete.

No authority, custody, withdrawal, or admin surface is added.

## Validation

- Exact-parent public LiteSVM RED against engine `d3133a93cf2cf219b69e0072cb4079d24a07f4f6`: `48,000 -> 39,576`
- Fixed public LiteSVM GREEN: prior claim remains `48,000`; terminal SPL payout is exactly `1,048,000`
- Engine all-target test suite passes
- `cargo kani --tests --only-codegen --features fuzz` passes

The end-to-end public-program regression is in the paired stacked `percolator-prog` PR.

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -15740,10 +15740,10 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
         &mut self,
         account: &mut PortfolioV16ViewMut<'_>,
         asset_index: usize,
-    ) -> V16Result<(u128, u128, u128, u128)> {
+    ) -> V16Result<(u128, u128, u128)> {
         let Some(leg_slot) = Self::active_leg_slot_for_asset(&account.as_view(), asset_index)?
         else {
-            return Ok((0, 0, 0, 0));
+            return Ok((0, 0, 0));
         };
         let mut leg = account.header.legs[leg_slot].try_to_runtime()?;
         let (k_now, f_now) = self.kf_target_for_leg(asset_index, leg)?;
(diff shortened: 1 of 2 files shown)


aeyakovenko:

PR #169, opened 2026-07-29
[BLOCKER LoF] Reconcile backing expiry during Recovery forfeit

aeyakovenko:
## Impact

A signed `ForfeitRecoveryLeg` transaction retained before a backing bucket expired could be landed after expiry. The Recovery capitalization path still treated the lapsed source lien as valid, allowing a claimant to consume an external provider's expired backing principal and withdraw it.

Public program reachability and the corresponding SBF regression are covered by aeyakovenko/percolator-prog#374.

## Root cause

The Recovery forfeit transition had no authenticated execution-slot input and did not reconcile source-bucket expiry before capitalizing a detached leg's claim. A cached pre-expiry market slot therefore remained sufficient after the transaction's actual execution slot crossed expiry.

## Change

- Add a slot-authenticated Recovery forfeit entry point.
- Reconcile expiry only for the selected source domain.
- Relabel the selected account's lapsed counterparty lien as impaired before capitalization.
- Cap utilization fees at the bucket's expiry slot.
- Preserve historical impaired claim floors while burning the current episode suffix when expired backing cannot lawfully fund it.
- Keep the existing API for callers that do not yet supply an authenticated slot.

This adds no custody path, admin authority, or DAO authority over user/provider funds.

## Validation

- Exact-base RED regression: `v16_recovery_forfeit_at_slot_impairs_expired_backing_and_preserves_floor`.
- Fixed engine suite: 50 unit + 9 backing fuzz + 2 resolved fuzz + 75 spec tests (136 total).
- Kani closure proof: `closure_recovery_expiry_impairment_preserves_claim_and_effective_totals`, 326 checks, 0 failures.
- `cargo fmt --all -- --check` and `git diff --check`.

Stacked on #168 because the test exercises the retained Recovery claim-floor semantics introduced there.

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -717,6 +717,94 @@ impl V16Core {
         Ok((required_face_num, required_backing_num))
     }
 
+    fn crystallize_source_lien_fee_for_effective(
+        source: &mut PortfolioSourceDomainV16Account,
+        impaired_effective: u128,
+    ) -> V16Result<u128> {
+        let live_effective = source.source_lien_effective_reserved.get();
+        if impaired_effective > live_effective {
+            return Err(V16Error::CounterUnderflow);
+        }
+        let live_fee = source.source_lien_capital_at_risk_fee_revenue.get();
+        let impaired_fee = if impaired_effective == live_effective {
+            live_fee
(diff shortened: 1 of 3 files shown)


aeyakovenko:

PR #173, opened 2026-08-06
[PARTIAL DoS] Make lapsed live source backing publicly progress

aeyakovenko:
## What changed

- Select the first lapsed source-backing domain referenced by a live portfolio during permissionless account refresh.
- Expire exactly one domain per auto-crank call, then return an explicit progress outcome before valuation.
- Keep the scan bounded by `PORTFOLIO_SOURCE_DOMAIN_CAP` and preserve unrelated/future domains.
- Add full-width expiry accounting and bounded-drain Kani proofs plus engine regression tests.

## Root cause and impact

A funded live portfolio with positive PnL sourced from a backing bucket could outlive that bucket expiry. The sole public auto-crank then reached valuation with stale source backing and returned `Stale` without a public transition that normalized the dependency. Unless a privileged authenticated update happened to refresh the route, the owner risk-reduction path could remain unavailable.

The fix makes that stale dependency itself a permissionless bounded continuation. Repeated calls expire the finite referenced-domain set, certify the account, and restore the owner reduction path. Expiry is value-neutral at the vault boundary and only reclassifies the affected backing/lien state.

## Public reproducer

Wrapper PR #135 pins this commit and adds a real LiteSVM sequence using public instructions: establish source backing, create funded exposure, advance beyond expiry, auto-crank one expiry step, auto-crank certification, then reduce exposure. It also promotes the prior quarantined blocker seed into the progressing corpus.

## Verification

- `cargo test --all-targets`: pass
- Kani expiry proof: 172 assertions and 4 covers pass
- Kani bounded scan proof: 83 assertions and 6 covers pass
- Wrapper exact-pin suite: 668 CU tests, 80 deterministic/fuzz regressions, 117 stateful tests, and 41 Kani harnesses pass
- `cargo fmt --all -- --check`: pass

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -2360,6 +2360,66 @@ impl V16Core {
         Ok((bucket, source))
     }
 
+    fn prepare_counterparty_backing_expiry_delta(
+        mut bucket: BackingBucketV16,
+        mut source: SourceCreditStateV16,
+        now_slot: u64,
+    ) -> V16Result<(BackingBucketV16, SourceCreditStateV16)> {
+        if bucket.status != BackingBucketStatusV16::Fresh || now_slot < bucket.expiry_slot {
+            return Err(V16Error::Stale);
+        }
+        let expired_unliened = bucket.fresh_unliened_backing_num;
+        let expired_liened = bucket.valid_liened_backing_num;
+        let expired_total = expired_unliened
+            .checked_add(expired_liened)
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #173 (c09d4575144f5be5fd0e8564d72fa22665a7e5bc) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #174, opened 2026-08-08
[REAL DoS] Preserve flat trade residual progress

aeyakovenko:
## What changed

- preserve asset-local close attribution when an ordinary final-leg trade leaves a uniquely attributable residual deficit
- make pending close ledgers a production-reachable `AdvanceClose` auto-crank class
- book one bounded residual chunk per crank
- keep completed and unattributed account deficits scope-local: they cannot force unrelated live markets into Recovery
- leave terminal market settlement to the explicit market-level resolution policy when provenance is ambiguous or account-local booking is complete
- cover single, batch, multi-chunk selector priority, observation independence, terminal payout, and multi-asset attribution behavior

## Root cause

The proof selector modeled `pending_close`, but the production classifier hard-coded that class to false and the production dispatcher had no close-advance arm. An ordinary matched final-leg close could therefore consume all loser principal, clear the last leg, and leave negative account PnL under a sticky market-wide bankruptcy lock with no successful rank-decreasing crank.

## User impact

An uncooperative underfunded trader could reach the state through normal trade APIs. Honest permissionless cranks remained successful no-ops while funded winners could not convert or withdraw released PnL. The fix keeps the risk-reducing trade available, durably attributes a uniquely attributable residual, and provides one bounded public booking step without granting the affected account authority to terminate unrelated markets. The configured market-level resolution route remains the terminal fallback.

## Validation

- `cargo test` (50 library, 11 backing, 2 resolved, 83 v16 specification tests)
- `cargo kani --tests --features fuzz --harness proof_v16_auto_crank_pending_close_priority_is_total --harness proof_v16_auto_crank_refresh_is_unique_observation_requiring_plan` (2/2)
- wrapper PR 135 public LiteSVM route matrix covers single/batch CPI/no-CPI, account-local booking, scope locality, permissionless stale resolution, terminal payout, and exact SPL conservation
- `git diff --check`

Base is the exact engine integration commit pinned by wrapper PR 135 (`9ffc4749a4b7e486f814090c7b43fb01a6df5dcf`).

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -171,8 +171,9 @@ pub fn v16_domain_pair_for_asset_index(asset_index: usize) -> V16Result<(usize, 
 /// account leg from which to choose a committed asset, so it must accrue a NEW
 /// price. `RefreshAccount { asset_index: Some(_) }` and EVERY other plan are
 /// dispatchable from committed on-chain state alone — `SettleBChunk` ignores
-/// price, `Liquidate` reads the current health cert, `DeclareRecovery` /
-/// `FinalizeRecovery` / `CloseResolved` / `NoAction` take no price — so a keeper
+/// price, `Liquidate` reads the current health cert, and `AdvanceClose` /
+/// `DeclareRecovery` / `FinalizeRecovery` / `CloseResolved` / `NoAction` take no
+/// price — so a keeper
 /// holding no fresh observation can still drive the account forward (no liveness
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #174 (94abbd2401b2f151243965c625db4ef39c0f513b) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #177, opened 2026-08-19
[REAL DoS] Release cured counterparty obligations

aeyakovenko:
## What changed

Auto-crank now classifies a released, zero-basis pending-loss obligation as actionable after its originating domain barrier is gone. If its K/F/B snapshots are current, the selected refresh step detaches it through the existing checked `clear_leg` transition; otherwise normal refresh settles it first.

## Root cause and impact

A public trade/cure sequence can cancel a bankrupt close while leaving the winning counterparty with a current zero-basis leg, nonzero loss weight, and one pending obligation. The sole public auto-crank previously returned `NoAction`, so repeated permissionless cranks committed byte-identical no-ops and the funded winner could not withdraw. This is an account-scoped persistent DoS reachable from normal public transitions.

This PR targets the `0a23b5f5fc85ddb0223089c66c29cbf1600be62b` engine lineage currently consumed by `percolator-prog` PR 135.

## Verification

- `cargo test`: 150/150 tests passed, including a bilateral long/short regression.
- Kani `proof_v16_released_flat_pending_obligation_is_publicly_actionable`: 0/4,199 checks failed; 4/4 covers reached.
- Kani `proof_v16_released_flat_pending_obligation_detach_is_value_neutral_progress`: 0/4,245 checks failed; 2/2 covers reached.
- A public LiteSVM TDD route in `percolator-prog` PR 135 reproduces the pre-fix lock and will be pinned to this commit for end-to-end validation.

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -12649,7 +12649,13 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
         account: &PortfolioV16View<'_>,
     ) -> V16Result<(
         ActionableSummaryV16,
-        (Option<usize>, Option<usize>, Option<usize>, Option<usize>),
+        (
+            Option<usize>,
+            Option<usize>,
+            Option<usize>,
+            Option<usize>,
+            Option<usize>,
+        ),
     )> {
         let mode = decode_market_mode(self.header.mode)?;
         let live = mode == MarketModeV16::Live;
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #177 (7219591416fe15496d2b043b7825aac622585522) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #178, opened 2026-08-19
[REAL LoF] Make oracle accrual cadence partition-invariant

aeyakovenko:
## Problem
A caller could partition identical elapsed time into many permissionless cranks and change the effective-price endpoint because the movement cap compounded from the current price. Delayed accrual also sampled funding only at the endpoint, so transaction cadence changed value transferred between counterparties.

## Fix
- add a bounded canonical one-slot accrual path (maximum 32 steps per call)
- anchor movement linearly for the lifetime of one raw target
- carry sub-basis-point movement residue across calls
- integrate funding per canonical logical step
- reject caller-supplied paths that differ from the unique engine-computed price/remainder prefix

The wrapper persists the carry and constructs the authenticated funding path in PR 135.

## Verification
- `cargo test`: 155/155
- `cargo kani --harness proof_v16_canonical_accrual_path_is_partition_invariant`: pass, 3/3 covers
- engine eager/delayed, irregular-partition, fractional movement, reset, and maximum-prefix regressions pass
- wrapper PR 135: 91/91 public, 145/145 stateful, 753/753 LiteSVM/CU, 131/131 Kani


aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #178 (149cfe56f339599be55223f42412fb8de8abe494) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #179, opened 2026-08-20
Make lapsed source backing auto-crank actionable

aeyakovenko:
## What changed

- classify account-local lapsed source backing as refresh work even when the health certificate epochs remain current
- prove the classifier predicate over every Boolean input combination
- add a production-state regression where unchanged price and funding preserve the certificate across expiry

## Root cause

Backing expiry is clock-driven, but the auto-crank classifier previously relied on unrelated certificate epochs to select `RefreshAccount`. At an unchanged price and funding point, those epochs can stay current after the backing bucket lapses, so the proven bounded expiry continuation existed but was unreachable from the public selector.

## Impact

The self-classifying crank can now reach the existing one-domain-per-call expiry transition without caller-selected economic state. This closes the composition gap needed by the wrapper INV-063 public expiry-boundary regression.

## Validation

- `cargo test` (162 runtime tests)
- `cargo test --test v16_spec_tests` (98/98)
- `cargo kani --tests --features fuzz --harness proof_v16_lapsed_source_backing_forces_live_refresh_even_with_current_cert`
- `git diff --check`


aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #179 (44847fd5ccf76a97dd31bb2c9dbd49ec94093d35) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #180, opened 2026-08-20
[PARTIAL DoS] Keep reset obligations crankable in Recovery

aeyakovenko:
## Problem

A public unilateral reduction can leave a prior-epoch `ResetPending` leg. If asset shutdown lands before its cleanup crank, the asset enters Recovery while preserving that reset obligation. The auto-crank classifier marks the account stale/actionable, but selects no refresh asset and dispatches ordinary accrual against Recovery, returning `LockActive`. Owner forfeit and delayed force-close still provide exits, so this is a partial crankability DoS and a contradiction of the selector totality claim, not a permanent funds lock.

## Fix

- use the reset-obligation asset as the refresh fallback when no accrual-eligible asset exists
- treat detachment of the selected leg during refresh as completed progress and skip post-refresh Recovery accrual
- add a direct production-body regression and two exhaustive selector/accrual Kani kernels

## Verification

- `cargo test --quiet`
- `cargo test --quiet --test v16_spec_tests v16_auto_crank` (16/16)
- `cargo kani --tests --features fuzz --harness proof_v16_auto_crank_refresh_target_includes_recovery_reset_obligation`
- `cargo kani --tests --features fuzz --harness proof_v16_detached_refresh_leg_skips_post_refresh_accrual`
- public LiteSVM composition is owned by wrapper PR 135

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -1880,6 +1880,20 @@ impl V16Core {
         !already_cleared && prior_reset_obligation && !pending_close_residual
     }
 
+    fn kernel_auto_crank_refresh_asset(
+        refresh_asset: Option<usize>,
+        reset_obligation_asset: Option<usize>,
+    ) -> Option<usize> {
+        refresh_asset.or(reset_obligation_asset)
+    }
+
+    fn kernel_refresh_detached_selected_leg(
+        selected_leg_before: bool,
+        selected_leg_after: bool,
+    ) -> bool {
+        selected_leg_before && !selected_leg_after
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #180 (7387e7a9c1aa1dbd337dc91e50ccfc11ce5109b2) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #181, opened 2026-08-21
[REAL LoF] Preserve terminal claims across source haircuts

aeyakovenko:
## What changed

- Separate elective live source-credit conversion from mandatory resolved settlement.
- At terminal close, debit positive PnL only by atoms actually converted to capital and demote the remaining source-attributed face into the ordinary terminal receipt pool.
- Add a randomized terminal value-partition regression plus full-width plain and function-contract Kani proofs.

## Root cause

Resolved close reused the live conversion rule and removed `face_burn` from PnL even when only `converted` atoms reached capital. At a tiny source-credit rate, a one-atom position could therefore erase a much larger unrelated terminal claim and leave the value ownerless in the vault.

## Validation

- `cargo test`
- `cargo test --features fuzz`
- `cargo kani --tests --features fuzz --harness proof_v16_terminal_source_haircut_retains_junior_face`
- `cargo kani --tests --features fuzz,contracts -Z function-contracts --harness contract_check_kernel_released_pnl_conversion_partition`
- Wrapper PR135 public LiteSVM certification across single/batch CPI/no-CPI routes


aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #181 (3b76b7948700ecc69fef69c84e7f5222c8a8a686) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #182, opened 2026-08-23
[PARTIAL LoF] Prevent split-account margin under-reservation

aeyakovenko:
## Finding

An independently generated public LiteSVM route showed that splitting one proportional risk increase across two portfolios reserved one fewer quote atom of source-credit lien than the aggregate trade. The engine ceiled risk notional but then floored each portfolio's margin requirement, so partitioning could reduce aggregate admission and backing requirements.

This is publicly reachable and violates split/merge conservatism. The demonstrated advantage is one quote atom per additional portfolio; the repro does not by itself establish a full vault drain, so this is labeled partial LoF.

## Fix

- ceil proportional initial and maintenance requirements
- use the same canonical requirement in liquidation planning and post-state certification
- align config-envelope validation and the v16.8 spec
- add a deterministic regression, randomized deployed-arithmetic property, and division-axiom Kani composition theorem

## Verification

- `cargo test`: 181/181
- `cargo test --features fuzz`: 226/226
- focused Kani: 0/21 failures, 3/3 covers
- independent public LiteSVM red/green coverage is retained in wrapper PR 135

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -194,7 +194,7 @@ price_funding_loss_X  = ceil(X * loss_budget_num / (10_000 * FUNDING_DEN))
 worst_liq_notional_X  = ceil(X * (10_000 + price_budget_bps) / 10_000)
 liq_fee_raw_X         = ceil(worst_liq_notional_X * cfg_liquidation_fee_bps / 10_000)
 liq_fee_X             = min(max(liq_fee_raw_X, cfg_min_liquidation_abs), cfg_liquidation_fee_cap)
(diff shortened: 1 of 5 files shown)


aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #182 (ba7a84b774870c3db9dd14d9747e8326b31f3a94) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #183, opened 2026-08-23
[REAL LoF] Preserve post-ADL effective position accounting

aeyakovenko:
## What changed

- define public position deltas after quantity ADL in effective-quantity units
- map same-side effective reductions back to retained raw K/F basis with a checked inverse
- use effective quantity consistently for OI, health, liquidation, full-close detection, recovery obligations, and resolved close
- retain raw basis only for K/F settlement and social-loss weight accounting

## Root cause

After quantity ADL, a portfolio leg retains its pre-ADL raw basis while aggregate OI is reduced. Trade, health, liquidation, and clear paths still treated that stale raw basis as live exposure. A public bilateral trade could therefore reduce more than the account's effective exposure while remaining below its raw basis. The maintained OI then diverged from the live legs, and later price/funding accrual was no longer zero-sum.

The wrapper LiteSVM regression reaches this from ordinary market setup, trades, and permissionless liquidation with no program-state injection. It reproduced on all four public trade routes.

## Validation

- `CARGO_INCREMENTAL=0 cargo test`
- `CARGO_INCREMENTAL=0 cargo test --features fuzz`
- `CARGO_INCREMENTAL=0 cargo build-sbf --tools-version v1.52`
- Kani inverse proof: 510 checks, 2/2 nonvacuity covers
- Kani resize contract: 1,432 checks
- Kani liquidation selector proof: 506 checks, 2/2 covers
- Kani clear and pending-obligation contracts

The wrapper public-interface red/green regression and route/CU controls are being retained in percolator-prog PR 135 against this immutable engine commit.


aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #183 (78c73bc8f6a7703374d7ca68f9c84ec8293209bc) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #184, opened 2026-08-23
[REAL DoS] Keep cross-domain source losses crankable

aeyakovenko:
## Bug\nA publicly reachable multi-asset account can hold source-attributed positive PnL in several domains whose fractional realizable backing sums to a whole atom. Health/loss estimation aggregated those fractions before atom rounding, while consumption rounded each domain independently. A permissionless refresh then demanded an unconsumable exact atom and returned LockActive, so every canonical crank rolled back.\n\n## Fix\n- round realizable support independently per source domain through one backing-capped helper\n- let loss settlement consume the whole-atom support actually available instead of requiring an overestimated exact amount\n- preserve fail-closed exact consumption for conversion routes\n\n## Evidence\n- valid-state engine transition regression for two half-atom domains\n- per-domain rounding fuzz regression\n- focused Kani proof\n- public LiteSVM regression lives on wrapper PR #135 and reaches the failure with three authenticated mark pushes\n\n## Verification\n- cargo test: 184/184\n- cargo test --features fuzz: 233/233\n- focused Kani: 1/1\n- wrapper SBF build + focused public LiteSVM regression: green

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -3932,20 +3932,29 @@ impl V16Core {
         Ok((reservation, source))
     }
 
+    // Source domains settle whole quote atoms independently. Round before a
+    // caller aggregates domains so fractional backing cannot become phantom support.
+    fn source_credit_state_realizable_support_for_claim_num(
+        state: SourceCreditStateV16,
+        claim_num: u128,
+    ) -> V16Result<u128> {
+        if claim_num == 0 || state.positive_claim_bound_num == 0 {
+            return Ok(0);
+        }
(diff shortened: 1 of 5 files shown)


aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #184 (592d538c29554d576a660943d5e09c23bcd2a2e0) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #185, opened 2026-08-24
[REAL DoS] Release flat source liens through auto-crank

aeyakovenko:
A publicly reachable flat portfolio can retain a source-credit lien after its backing provider withdraws. The existing selector reports NoAction, while conversion and close remain blocked, leaving funded positive PnL without a bounded exit.

This adds a bounded, observation-free ReleaseSourceLiens auto-crank class, dispatches the existing validated release transition in production, and proves selector totality/priority. The wrapper LiteSVM public-route regression is being added to percolator-prog PR 135 against this exact commit.

Verified: cargo test (185), cargo test --features fuzz (234), the new Kani selector proof, observation predicate proof, and selector/progress/summary function contracts.

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -192,6 +192,7 @@ pub fn auto_crank_plan_requires_caller_observation(plan: &AutoCrankPlanV16) -> b
         AutoCrankPlanV16::AdvanceClose
         | AutoCrankPlanV16::SettleBChunk { .. }
         | AutoCrankPlanV16::Liquidate { .. }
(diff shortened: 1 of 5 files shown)


aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #185 (fdf11670d1133ade1647aac8280356cea1536b78) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #186, opened 2026-08-24
[REAL DoS/CU] Chunk maximum-shape source-lien release

aeyakovenko:
## Public impact
A portfolio can publicly accumulate all 28 wrapper-supported source domains and retain a real lien in each. After it becomes flat, the sole automatic crank selects ReleaseSourceLiens, but the all-at-once mutation exhausts the 1.4M SVM CU meter. The funded positive claim then cannot convert or close in a live market.

## Fix
Release one source-lien domain per automatic crank. Every successful call strictly decreases the finite lien-domain count, and the selector repeats the same observation-free continuation until no lien remains.

## Evidence
- Engine regression proves two real source liens progress 2 -> 1 -> 0 before exact conversion.
- Full engine base suite: 186 passing.
- Full engine fuzz/reference suite: 235 passing.
- The stacked wrapper LiteSVM regression constructs all 28 liens using only public instructions and measures every release call below the transaction limit.

Stacked on engine PR #185 because that PR introduces the ReleaseSourceLiens continuation.


aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #186 (b10b3454dd03dcf4c04a020dc1a90381ff179200) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #187, opened 2026-08-28
[REAL LoF] Preserve domain attribution across multi-asset loss detach

aeyakovenko:
Stacked on #186.

A finding-blind public LiteSVM trace on the parent pin opens two assets, realizes a loss, detaches the loss-bearing leg, and then invokes the sole automatic crank. Before this change, the surviving asset was treated as the loss source: 100,100 atoms of unrelated-domain insurance were spent and the attacker-controlled coalition extracted 95,500 atoms of profit.

The fix reuses the existing `liquidation_lock` field. Unattributed multi-asset loss remains locked across detach; automatic liquidation remains permissionless and risk-reducing, but cannot spend insurance, book B, assign explicit loss, or charge a fee to an arbitrary surviving domain. No layout or API changes are introduced.

Verification:
- engine `cargo test --features fuzz`: 129/129
- existing uncovered-loss Kani theorem: 0/103 failed, 2/2 covers
- sticky-lock lifecycle Kani theorem: 0/7 failed, 3/3 covers
- wrapper PR #135 exact SBF: 101/101 regressions, 205/205 stateful, 886/886 CU, 174/174 wrapper Kani
- fixed public trace: one strict risk-reducing crank, zero foreign insurance spend, zero coalition profit, exact no-progress rollback, bounded exit, exact SPL supply reconciliation

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -314,6 +314,15 @@ fn uncovered_loss_remains_with_open_risk(
     uncovered_loss_after_principal != 0 && !active_bitmap_is_empty(remaining_active_bitmap)
 }
 
+#[inline]
+fn unattributed_loss_lock_after_pnl(
+    was_locked: bool,
+    active_bitmap: V16ActiveBitmap,
+    new_pnl: i128,
+) -> bool {
+    new_pnl < 0 && (was_locked || active_bitmap_count_ones(active_bitmap) > 1)
+}
+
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #187 (9b737fdcec16f3709c0651f4ecc7488b4917f2d8) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #188, opened 2026-08-29
[REAL LoF] Preserve source-local backing attribution

aeyakovenko:
## Impact
A portfolio with equal claims in two source domains could consume backing from the funded domain while the generic PnL update burned the first, unfunded claim. The remaining funded claim could then consume the same backing a second time and become withdrawable value.

## Fix
- burn each source claim in the same bounded loop that consumes that source backing
- carry the preburned face into the canonical PnL update
- block conversion while source liens remain so non-atomic engine callers cannot observe partial mutation
- add exact-once runtime coverage and symbolic proofs for both flat source liens and active source exposure

No layout, API, or persisted-state change.

## Verification
- `cargo test --all-targets`: 187/187
- `proof_v16_flat_source_claim_lien_blocks_convert`: pass, cover satisfied
- `proof_v16_open_source_claim_exposure_blocks_convert`: pass, cover satisfied
- wrapper fixed-seed and generated LiteSVM invariant coverage passes on exact pin `422893fa`

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -17755,7 +17755,10 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
             return Ok(0);
         }
         let has_source_claims = Self::account_has_source_claims(&account.as_view())?;
-        if has_source_claims && self.account_has_active_source_claim_exposure(&account.as_view())? {
+        if has_source_claims
+            && (Self::account_has_source_liens(&account.as_view())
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #188 (422893fa0a43a2d07697e88dd2009dcef8dce1d5) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #190, opened 2026-08-29
Canonicalize source-domain allocation order

aeyakovenko:
## Summary
- canonicalize persisted source-domain entries by domain before economic scans
- preserve the existing linear fast path for already ordered state
- add native and Kani regressions for field-preserving canonicalization

## Public-route evidence
A wrapper LiteSVM regression with two source domains and asymmetric backing fees showed that reversing otherwise identical signed trade history changed which domain funded the lien, moving 2,378 quote atoms between the target payout and provider earnings. The engine change makes the normalized economics identical for both orders.

This is classified as deterministic-allocation correctness under INV-041, not public theft or persistent DoS: both pre-fix outcomes remained within signed fee bounds and the selected provider supplied the backing.

## Verification
- cargo test: 50 lib, 13 backing, 2 resolved, 123 v16 spec tests
- focused Kani proof: 1/1, 3,045 checks, cover satisfied


aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #190 (c0dec8ce1e0f193d428f33f413c3116139fede55) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #191, opened 2026-08-29
[REAL DoS] Restore live progress for impaired source liens

aeyakovenko:
## Counterexample

A public trade-created flat account can retain a live source-claim lien after one backing bucket expires while a sibling source remains fresh. After authenticated expiry commits, the expired bucket is `Impaired`, so the lapsed-Fresh scan no longer finds it; the old release predicate requires every lien to be fresh, so auto-crank reports no action. The fresh sibling claim and provider encumbrance remain blocked indefinitely in a Live market.

This is stacked on the canonical source-domain-order change because the public two-domain invariant trace that found it depends on deterministic source ordering.

## Fix

- classify one bounded flat source-lien normalization from production state
- make authenticated backing expiry preempt any release at the same slot
- crystallize an impaired counterparty lien into the account impaired-claim lane
- retire the matching market impaired encumbrance
- release independently fresh sibling liens on subsequent bounded cranks
- keep the public crank API and persisted layout unchanged

## Verification

- `cargo test --all-targets`: 189 passed
- `proof_v16_flat_source_lien_normalization_is_total_and_fail_closed`: 24/24 assertions, 5/5 covers
- `proof_v16_lapsed_source_backing_preempts_flat_lien_normalization`: 2/2 assertions, 3/3 covers
- native finite-sequence regression proves expiry, two one-domain normalization steps, retained impaired attribution, and conversion of the fresh sibling claim

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ Source-checkable counts in this checkout:
 | --- | ---: |
 | Plain Kani proofs in `tests/proofs_v16.rs` | 289 |
 | Plain Kani proofs in `tests/proofs_v16_arithmetic.rs` | 11 |
-| Plain Kani proofs in `src/v16_proofs.rs` | 38 |
+| Plain Kani proofs in `src/v16_proofs.rs` | 40 |
 | Function-contract proofs in `src/v16_proofs.rs` | 51 |
 | Production `kernel_*` helpers in `src/v16.rs` | 17 |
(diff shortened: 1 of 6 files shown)


aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #191 (d604ca09b7e584d3875ce4516bab1186346bf4a6) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #192, opened 2026-08-30
[BLOCKER LoF] Preserve source claims across crank and leg order

aeyakovenko:
## Confirmed public violations

This PR fixes two order-dependent source-accounting defects reached by ordinary public wrapper instructions, without state injection.

1. **Positive-face overburn.** When a negative K/F delta reached an account with positive source-attributed PnL, any uncovered tail burned the entire prior positive face and then subtracted the tail again. On the parent engine, the same public lifecycle paid the winner 3,100 or 3,300 quote atoms solely according to permissionless crank order.
2. **Whole-account K/F ordering.** Refresh previously settled active legs in persisted slot order. A newly observed gain could therefore erase a newly observed loss before that loss reserved its source-domain backing. Both keeper-selected account order and portfolio leg-slot order could change terminal source attribution.

The PR135 LiteSVM regression crosses all four account-order and leg-slot-order combinations. On this head every combination ends with the same exact state: winner withdrawal 3,300, counterparty withdrawal 4,800 plus a 50 terminal payout, and zero terminal vault, C_tot, or SPL residue.

## Fix

- Burn retained positive face one-for-one with the actual post-support loss instead of erasing it and charging the uncovered tail twice.
- Build one bounded K/F plan from all active legs before applying any newly observed delta.
- Apply negative entries before nonnegative entries, then order by source domain and canonical leg slot.
- Keep source accounting domain-local; the plan orders settlement but does not net value across domains.
- Cache expensive K/F arithmetic for CU, then reload the live K/F target and revalidate target, phase, and source domain immediately before mutation.
- Update the normative specification to state the same rule.

There is no public API or persisted-layout change.

## Verification

- Engine runtime/property/spec suites: 239/239 pass with `--features fuzz`.
- `proof_v16_unbacked_loss_burns_positive_face_one_for_one`: 0/135 failures; 3/3 covers.
- `proof_account_kf_settlement_key_roundtrip_and_priority`: 0/115 failures.
- `proof_account_kf_settlement_plan_insert_preserves_order_and_multiset`: 0/144 failures.
- Wrapper PR135: 101/101 public regressions, 207/207 stateful tests, 916/916 LiteSVM/CU tests, and 193/193 Kani harnesses.
- Maximum-shape measurements on the exact final SBF: 14-feed refresh 931,087 CU under its unchanged 950,000 guardrail; 14-leg/28-source-domain exit 933,246 CU under the 1.4M transaction limit.

Final engine SHA: `a6e3c79f2d6c3afdfb82260951d8a5be85f8fa5d`.


aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #192 (a6e3c79f2d6c3afdfb82260951d8a5be85f8fa5d) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #193, opened 2026-08-30
[PARTIAL DoS] Clamp Recovery pair work to effective OI

aeyakovenko:
## Finding

Recovery pair force-close work is caller supplied and may be stale by landing time. The wrapper previously bounded it with retained raw basis, but after quantity ADL the executable quantity is lower effective OI. When both side ADL indices are non-unit, effective + 1 reaches the engine and rejects instead of landing the maximal valid close.

This is a permissionless-progress reliability violation, not a proven permanent funds lock: an exact fresh quantity still succeeds.

## Fix

Add force_close_recovery_pair_not_atomic, which treats the requested quantity as a work budget, clamps it to both canonical effective leg quantities and both OI lanes, then dispatches the matched reduction in the correct signed orientation.

## Verification

- New engine regression constructs both non-unit ADL sides through normal trades and owner reductions.
- effective + 1 lands exactly effective and clears both OI lanes.
- CARGO_INCREMENTAL=0 cargo test: 191 tests passed.

Stacked on the current source-domain branch because the wrapper invariant branch is pinned to that engine lineage.

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -17448,6 +17448,62 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
         Ok(())
     }
 
+    /// Closes one opposite-side pair in Recovery while treating caller quantity as a work budget.
+    /// The canonical effective quantities, rather than retained raw bases, bound the landed close.
+    pub fn force_close_recovery_pair_not_atomic(
+        &mut self,
+        account_a: &mut PortfolioV16ViewMut<'_>,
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #193 (ce590d9bd3e8f55eaf2e0321f36ef11fbc003d26) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #194, opened 2026-08-31
[PARTIAL DoS] Restore bounded terminal cleanup after backing expiry

aeyakovenko:
## Finding

A public wrapper lifecycle can finish every user claim and dematerialize every portfolio while retaining expired provider backing, historical insurance spend, and claim-free custody. The parent engine has no bounded transition that normalizes those stocks, so `CloseSlab` remains permanently blocked. This is a terminal market-teardown and unattributed-custody DoS, not trader-principal loss.

## Fix

- advance terminal cleanup in bounded 256-asset steps
- stop at live Fresh backing; retry at the parked frontier errors until withdrawal or authenticated expiry
- expire elapsed backing and recredit paired historical insurance before retirement
- retire only the final claim-free residual
- expose an explicit outcome so the wrapper can persist progress and pair final retirement with custody burn

The cursor is only an asset index and remains valid across slot changes because the scan never advances beyond still-live Fresh backing.

## Public evidence

PR135 reproduces the parent lock through public trade, mark, liquidation, resolution, payout, provider-withdrawal, insurance-withdrawal, portfolio-close, and slab-close routes across all four single/batch CPI/no-CPI opening transports. The fixed expired partition is exact: `751 = 1 provider withdrawal + 123 insurance withdrawal + 627 retired surplus`; the pre-expiry control is `751 = 751 provider withdrawal`. No program-owned bytes are injected in this lifecycle test.

The paired maximum-shape test lifts only the empty-slot dimension after public reachability: 5,782 assets complete in exactly 23 cross-slot calls, max 136,097 CU. Malformed custody accounts reject before engine mutation.

## Verification

- 196 engine host/runtime/spec tests
- six focused Kani proofs covering total priority, error-or-strict-progress, exact retirement, full framing, and paired-domain recredit
- PR135: 7 unit, 111 regression, 227 stateful/model, 917 LiteSVM/CU, and 193 Kani harnesses
- final SBF: 1,255,960 bytes, SHA-256 `d70956aa6559f9e3cdd5dde9553866227cea53b8df560e8d2da30b7f4252d837`


aeyakovenko:
Superseded by consolidated engine dependency #195. The exact head of #194 (6f3c5c124a68c1103a2ecd995ff4a10b3af247f8) is an ancestor of #195 head 495a5590c97055bd71c6f94d849ff0298f243145, so its complete change set is retained there.

PR #196, opened 2026-09-02
[PARTIAL DoS] Expire live backing at authenticated crank slot

aeyakovenko:
## Impact
A publicly reachable Live account can have source backing that is fresh at the committed market slot but lapsed at the authenticated execution slot. Auto-crank classifies the expiry as actionable at the authenticated slot, then the refresh path rechecks against the older committed slot and can reject every crank candidate instead of progressing. The owner exit remains live, so this is a crank-only partial DoS rather than a permanent funds lock.

## Fix
When the selected plan is RefreshAccount, expire the first account-scoped backing bucket using the same authenticated `work.now_slot` used by classification, before ordinary refresh/accrual. The call remains one bounded transition.

## TDD evidence
- Parent: `v16_auto_crank_expires_authenticated_slot_backing_before_live_refresh` fails because the first call returns AccountCurrent and leaves the lapsed bucket Fresh.
- Head: the first call returns SourceBackingExpired, and the next call performs ordinary catch-up.
- All 23 `v16_auto_crank_` spec tests pass.
- Wrapper PR135 has an independent public LiteSVM scenario that exposed the contradiction after its legacy failure quarantine was removed.

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -15145,6 +15145,24 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
                 AutoCrankOutcomeV16::Progressed(PermissionlessProgressOutcomeV16::AccountCurrent)
             }
             AutoCrankPlanV16::RefreshAccount { asset_index } => {
+                // Classification uses the authenticated execution slot, so dispatch must expire
+                // the same account-scoped prerequisite at that slot. Using header.current_slot
+                // inside the legacy refresh path can leave a newly lapsed bucket Fresh, after
+                // which K/F settlement rejects Stale and the selected plan cannot progress.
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Closing after the wrapper-level public replay disproved causality. The changed engine ordering is behaviorally cleaner, but the parent already makes real market progress on the first call, and the original public crank contradiction remains unchanged on this head. This therefore does not meet the project standard for a demonstrated LoF/DoS fix.

Gelettro:
When mainent  ?🤔😳

Le mer. 2 sept. 2026, 05:35, anatoly yakovenko ***@***.***> a
écrit :

PR #198, opened 2026-09-03
[REAL LoF] Retain pending obligations through Resolved close

aeyakovenko:
## Impact

A zero-basis, nonzero-loss-weight Recovery obligation could be closed first after the market entered Resolved, even while a real position remained on the opposite side. That deleted the pending-obligation count and loss weight before the opposite position settled. The later B allocation then used a smaller denominator, making close order redistribute payout between independent users.

The public wrapper LiteSVM reproduction on engine `495a5590c97055bd71c6f94d849ff0298f243145` observed attacker-first payouts `attacker=14, victim=10` versus debtor-first `attacker=12, victim=12`, while aggregate custody remained conserved. The early route also changed pending count `1 -> 0`, loss weight `40,000 -> 20,000`, and target B `1e17 -> 2e17`.

## RED on current master parent

Base: `8eb7142aada316f6c476f5c4fa815d3a806706d5`

```text
cargo test --test v16_spec_tests v16_resolved_close_retains_pending_obligation_until_opposite_position_is_gone -- --exact --nocapture

assertion failed
  left: Closed { payout: 1000 }
 right: ProgressOnly
```

The regression uses engine transitions to open a matched pair, force asset Recovery, forfeit one side into a zero-basis pending obligation, resolve the market, and attempt both resolved-close orders.

## Root cause

`detach_solvent_active_legs_for_resolved_close` treated every `close_q == 0` leg as either reset residue or an already-released obligation and called `clear_leg_at_slot_inner(..., Some(0))`. It did not call `recovery_pending_obligation_release_allowed`, despite the Recovery path using that predicate to retain loss weight until the opposite side has no real positions.

## Fix

For a zero-basis, nonzero-loss-weight leg, consult `recovery_pending_obligation_release_allowed` before detach. If release is not yet allowed, skip that leg rather than erroring or globally blocking the scan. A one-leg account returns `ProgressOnly`; after the opposite real position closes, the same obligation detaches and all pending/loss-weight counters reach zero.

## Verification

```text
cargo test --test v16_spec_tests v16_resolved_close_ -- --nocapture
# 5 passed

cargo test --test v16_spec_tests v16_recovery_forfeit_retains_loss_weight_until_opposite_positions_settle -- --exact
# 1 passed

cargo test --test v16_spec_tests v16_auto_crank_clears_released_recovery_obligation_with_finalized_close -- --exact
# 1 passed

cargo fmt --all -- --check
git diff --check
```

### DIFF
--- a/src/v16.rs
+++ b/src/v16.rs
@@ -19448,6 +19448,14 @@ impl<'a, T> MarketGroupV16ViewMut<'a, T> {
                 let account_effective_q = V16Core::effective_abs_quantity_for_leg(asset, leg)?;
                 let close_q = side_effective_oi_q.min(account_effective_q);
                 if close_q == 0 {
+                    if leg.basis_pos_q == 0
+                        && leg.loss_weight != 0
+                        && !self
+                            .recovery_pending_obligation_release_allowed(asset_index, leg.side)?
+                    {
+                        slot += 1;
+                        continue;
+                    }
                     // Zero-effective legs are either prior-reset residue or a
                     // released pending-loss obligation. They own no live OI,
                     // but still have to detach so resolved close can finish.
--- a/tests/v16_spec_tests.rs
+++ b/tests/v16_spec_tests.rs
@@ -7787,6 +7787,112 @@ fn v16_recovery_forfeit_retains_loss_weight_until_opposite_positions_settle() {
 }
 
 #[test]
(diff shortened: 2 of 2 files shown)


aeyakovenko:
Closing as superseded by #202. The pending-obligation resolved-close release guard is covered in the new engine branch, and wrapper PR aeyakovenko/percolator-prog#440 carries the public INV-076 regression for this behavior.

PR #201, opened 2026-09-13
Fix resolved mixed-role debt attribution

aeyakovenko:
## Summary
- preserve late resolved debt attribution after the creditor/source leg exits
- keep resolved mixed-role terminal claims backed by the original opposing economic debt until settlement

## Evidence
- Public-route wrapper regression in dependent percolator-prog branch reproduces the failure on the old engine pin: owner 2 receives 310000 instead of expected 340000 after mixed creditor/debtor close ordering.
- Engine fix updates the v16 attribution path only.

## Validation
- Validated by the dependent wrapper LiteSVM red/green selector and INV-039 surrounding regression suite in branch codex/astra-row419-attribution-fix-20260913.


aeyakovenko:
Closing as superseded by #202. #202 stacks this resolved mixed-role debt attribution fix with the newer resolved-source residual entitlement fix and is the engine branch consumed by wrapper PR aeyakovenko/percolator-prog#440.

